PR #2, opened 2018-08-21
Fix artifactId in README.md

ractive:


ynojima:
Thank you for pointing out! I'll merge this later.

PR #4, opened 2018-11-23
Fixing signature validation on AttestationStatements

hanndecl:
If the signature is valid an Validation exception is thrown. 
I added some tests in order to verify if everything went wel.


hanndecl:
I have also code for generating a selfsigned certificate with BouncyCastle. Do you want that I use this?
Another option is removing those tests

ynojima:
Thank you for pointing out! I'll merge this after I fix the cause that make a test case failed.

PR #5, opened 2019-01-13
remove WebAuthnAuthenticationContext#credentialId

mbakhoff:
WebAuthnAuthenticationContext has a credentialId field that was used up until commit 7a16cdc to find the associated authenticator data. Since commit 6943306 the field is unused, but still required and assigned in the constructor. This is confusing and may cause caller code to be more complicated.

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/response/WebAuthnAuthenticationContext.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/response/WebAuthnAuthenticationContext.java
@@ -31,13 +31,11 @@ public class WebAuthnAuthenticationContext extends AbstractWebAuthnContext {
     //~ Instance fields ================================================================================================
 
     // user inputs
-    private final byte[] credentialId;
     private final byte[] authenticatorData;
     private final byte[] signature;
 
     @SuppressWarnings("squid:S00107")
-    public WebAuthnAuthenticationContext(byte[] credentialId,
-                                         byte[] clientDataJSON,
+    public WebAuthnAuthenticationContext(byte[] clientDataJSON,
                                          byte[] authenticatorData,
                                          byte[] signature,
                                          String clientExtensionsJSON,
@@ -53,20 +51,17 @@ public class WebAuthnAuthenticationContext extends AbstractWebAuthnContext {
                 expectedExtensionIds
         );
 
-        this.credentialId = credentialId;
         this.signature = signature;
         this.authenticatorData = authenticatorData;
     }
 
-    public WebAuthnAuthenticationContext(byte[] credentialId,
-                                         byte[] clientDataJSON,
+    public WebAuthnAuthenticationContext(byte[] clientDataJSON,
                                          byte[] authenticatorData,
                                          byte[] signature,
                                          ServerProperty serverProperty,
                                          boolean userVerificationRequired
     ) {
         this(
-                credentialId,
                 clientDataJSON,
                 authenticatorData,
                 signature,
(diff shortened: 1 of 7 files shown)


ynojima:
Thank you for pointing out!
Your indication that credentialId is not currently used in assertion verification is reasonable.
However, since I'd like to add an extension point to accept custom validators in the future, and custom validators may need credentialId, so I would like to keep it.

mbakhoff:
Currently the WebAuthnAuthenticationContextValidator#validate method is passed both the authenticationContext and the authenticator (which contains the credentialId). Would the authenticator also be passed to the future custom validators? If that is the case, then it might not be necessary to keep a copy in authenticationContext.

ynojima:
I will pass the authenticator to the future custome validators, but since the positioning of WebAuthnAuthenticationContext and Authenticator are different, I would like to keep credentialId in WebAuthnAuthenticationContext.
authenticationContext is an envelope for inputs from client platform, on the other hand, authenticator is an envelope for data saved in registration process. Sorry for lack of documentation.

mbakhoff:
Thanks for looking into this and explaining. I'll close the PR. 

> authenticationContext is an envelope for inputs from client platform, on the other hand, authenticator is an envelope for data saved in registration process 

That's actually exactly how I understood it and that's why I was confused. The client doesn't return the crendentialId as a separate field (https://www.w3.org/TR/webauthn/#iface-authenticatorassertionresponse), so it's not obvious why it should be in the authenticationContext. Furthermore, the credentialId may be included inside the authenticatorData field. So right now the authenticationContext may contain two copies of the credentialId (if the AT flag is set): 

* authenticationContext.getCredentialId();
* authenticatorDataConverter.convert(authenticationContext.getAuthenticatorData()).getAttestedCredentialData().getCredentialId();

But in the end it's not that important. Thanks for the awesome library :)

ynojima:
> The client doesn't return the crendentialId as a separate field (https://www.w3.org/TR/webauthn/#iface-authenticatorassertionresponse), so it's not obvious why it should be in the authenticationContext. Furthermore, the credentialId may be included inside the authenticatorData field.

credentialId can be retrieved from `id` field of `PublicKeyCredential` interface(https://www.w3.org/TR/webauthn/#iface-pkcredential), which is a container for `AuthenticatorAssertionResponse` interface. You don't need to extract `AuthenticatorAssertionResponse` by yourself :)
Is that why you said "This is confusing and may cause caller code to be more complicated"?

FYI, 
> For attestation signatures, the authenticator MUST set the AT flag and include the attestedCredentialData. For authentication signatures, the AT flag MUST NOT be set and the attestedCredentialData MUST NOT be included.
https://www.w3.org/TR/webauthn/#sec-authenticator-data

So, authenticationContext never contains attestedCredentialData.

I hope this will help you.

mbakhoff:
> For authentication signatures, the AT flag MUST NOT be set and the attestedCredentialData MUST NOT be included.

Ah, I suspected that might be the case. Makes sense. 

> credentialId can be retrieved from id field of PublicKeyCredential interface

Right, that's true. I didn't notice that at first. Thanks for pointing it out. In this case it does make sense to include the rawId in the context. 

> Is that why you said "This is confusing and may cause caller code to be more complicated"?

This is what I was thinking: The constructor of authenticationContext requires the credentialId to be non-null, so it must be important. However credentialId should not be required to check the signature if the public key is already provided by the authenticator object. Did I misunderstand the protocol and the credentialId from the client is actually required to ensure security?

I now realize that the credentialId is not required to verify the signature, but it's useful for choosing the right authenticator (if the user has several).

ynojima:
Thank you for clarification.

> I now realize that the credentialId is not required to verify the signature, but it's useful for choosing the right authenticator (if the user has several).

To the best of my knowledge, that is correct.

Issue #48, opened 2019-03-07
enum of ALG_SIGN for Authentication Algorithm


tnorimat:
Related to #47, I think we need to prepare emun of ALG_SIGN like existing AttestationType.java.

Please refer to the following specification:
https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-registry-v2.0-rd-20180702.html#authentication-algorithms

ynojima:
This is to be addressed. Thank you for pointing out.
Could you fix it?

tnorimat:
Yes, I could.

ynojima:
Fixed by #52

Issue #42, opened 2019-03-07
emun of USER_VERIFY for User Verification Method


tnorimat:
Related to #41, I think we need to prepare emun of USER_VERIFY like existing AttestationType.java.

Please refer to the following specification:
https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-registry-v2.0-rd-20180702.html#user-verification-methods

ynojima:
Thank you for suggestion. This should be fixed as you proposed (introducing new enum type).

ynojima:
Fixed by https://github.com/webauthn4j/webauthn4j/commit/326c72345431ff6c91284fd03538523e1f61b2ce

PR #52, opened 2019-03-07
Issue #48 enum of ALG_SIGN for Authentication Algorithm

tnorimat:


webauthn4j-metadata/src/main/java/com/webauthn4j/metadata/data/statement/AuthenticationAlgorithm.java
```diff
@@ -0,0 +1,115 @@
...
+    ALG_SIGN_RSA_EMSA_PKCS1_SHA256_DER(0x0009),
+    ALG_SIGN_RSASSA_PSS_SHA384_RAW(0x000A),
+    ALG_SIGN_RSASSA_PSS_SHA512_RAW(0x000B),
+    ALG_SIGN_RSASSA_PKCSV15_SHA256_RAW(0x000C),
+    ALG_SIGN_RSASSA_PKCSV15_SHA384_RAW(0x000D),
+    ALG_SIGN_RSASSA_PKCSV15_SHA512_RAW(0x000E),
+    ALG_SIGN_RSASSA_PKCSV15_SHA1_RAW(0x000F),
+    ALG_SIGN_SECP384R1_ECDSA_SHA384_RAW(0x0010),
+    ALG_SIGN_SECP521R1_ECDSA_SHA512_RAW(0x0011),
+    ALG_SIGN_ED25519_EDDSA_SHA256_RAW(0x0012);
```
ynojima: Since it is not introduced as constants but as enum, "ALG_SIGN_" prefix is unnecessary for enum members.

tnorimat: Should I also remove USER_VERIFY_ prefix?
https://github.com/webauthn4j/webauthn4j/issues/42

FIDO Registry of Predefined Values tells that "The USER_VERIFY constants are flags in a bitfield represented as a 32 bit long integer. ", therefore, USER_VERIFY is also constants same as ALG_SIGN.

ynojima: Yes, I overlooked it by mistake.

tnorimat: Thank your for your comments.
BTW, should I remove ATTESTATION_ prefix from existing AttestationType.java ?
It seems that ATTESTATION is also constant mentioned in https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-registry-v2.0-rd-20180702.html#authenticator-attestation-types

ynojima: Good catch. It needs to be addressed too.

tnorimat: Hope I'll deal with it.

Issue #50, opened 2019-03-07
enum of ALG_KEY for Public Key Representation Format


tnorimat:
Related to #49, I think we need to prepare emun of ALG_KEY like existing AttestationType.java.

Please refer to the following specification:
https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-registry-v2.0-rd-20180702.html#public-key-representation-formats

ynojima:
This one need to be fixed too. Thanks!

ynojima:
Fixed by #53

PR #53, opened 2019-03-07
Issue #50 enum of ALG_KEY for Public Key Representation Format

tnorimat:


webauthn4j-metadata/src/main/java/com/webauthn4j/metadata/data/statement/PublicKeyRepresentationFormat.java
```diff
@@ -0,0 +1,76 @@
...
+ * The supported publik key representation format(s).
+ * See section 3.6.2 Public Key Representation Formats of FIDO Registry of Predefined Values.
+ */
+public enum PublicKeyRepresentationFormat {
+
+    ALG_KEY_ECC_X962_RAW(0x0100),
+    ALG_KEY_ECC_X962_DER(0x0101),
+    ALG_KEY_RSA_2048_RAW(0x0102),
+    ALG_KEY_RSA_2048_DER(0x0103),
+    ALG_KEY_COSE(0x0104);
```
ynojima: Since it is not introduced as constants but as enum, "ALG_KEY_" prefix is unnecessary for enum members.

PR #56, opened 2019-03-08
remove ATTESTATION_ prefix, and add reference to specification

tnorimat:


ynojima:
Thank you for the fix!

PR #62, opened 2019-03-12
PublicKeyCredentialRequestOptions.timeout is an optional value 

kvandermast:
PublicKeyCredentialRequestOptions.timeout is an optional value and can be overridden by the client. 

The object has been updated to better match the IDL and provide the ability to use it as a JSON object when preparing for the authentication flow.


ynojima:
Thank you for your pull-request! It seems two different issues are mixed in one pull-request. Please separate them from next time.
First commit seems reasonable and I merged it into master, but for second commit, I don't see enough reason to change the interface. If you believe it need to be changed, please create another pull-request with its reason.

Issue #63, opened 2019-03-14
cannot specify JSON Metadata Statement file paths to JsonFileMetadataItemsProvider


tnorimat:
JsonFileMetadataItemsProvider reads Metadate Statements from JSON files, but the user of this class cannot specify their paths.

ynojima:
Fixed by #64. Thank you!

PR #64, opened 2019-03-14
Issue #63 cannot specify JSON Metadata Statement file paths

tnorimat:


webauthn4j-metadata/src/test/java/com/webauthn4j/metadata/JsonFileMetadataItemsProviderTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void fetchMetadata() throws Exception {
+       List<Path> paths= new ArrayList<>(4);
+       paths.add(Paths.get(ClassLoader.getSystemResource("com/webauthn4j/metadata/JsonMetadataItem_fido2.json").toURI()));
+       paths.add(Paths.get(ClassLoader.getSystemResource("com/webauthn4j/metadata/JsonMetadataItem_u2f.json").toURI()));
+       paths.add(Paths.get(ClassLoader.getSystemResource("com/webauthn4j/metadata/JsonMetadataItem_uaf.json").toURI()));
+       paths.add(Paths.get(ClassLoader.getSystemResource("com/webauthn4j/metadata/JsonMetadataItem_unknown_protocol.json").toURI()));
+       JsonFileMetadataItemsProvider provider = new JsonFileMetadataItemsProvider(jsonConverter, paths);
+       readMetadataItem(provider.provide());
+       // read again
+       readMetadataItem(provider.provide());
```
ynojima: Please assert the same (cached) object is returned.

tnorimat: Would it mean that I need to check whether the object returned from second call of provider.provider() is the same one from first call in order to confirm that this test method run through all codes of JsonFileMetadataItemsProvider.provider() ?

ynojima: Yes

tnorimat: So sorry...

PR #68, opened 2019-03-18
Removed switch statement in favor of streams for the create-method.

kvandermast:
Removed the switch statements in the Algorithm-enums "create" methods in favor of streams.


webauthn4j-core/src/main/java/com/webauthn4j/data/attestation/statement/COSEAlgorithmIdentifier.java
```diff
@@ -39,34 +41,19 @@
...
-            case -7:
-                return ES256;
-            case -35:
-                return ES384;
-            case -36:
-                return ES512;
-            default:
-                throw new IllegalArgumentException("value '" + value + "' is out of range");
-        }
+    public static COSEAlgorithmIdentifier create(long value) {
```
ynojima: Thank you for pointing out!

PR #69, opened 2019-03-18
Generified AttestationStatementValidators

kvandermast:
Moved the "supports" method to a abstract class

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/validator/attestation/statement/AbstractStatementValidator.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/validator/attestation/statement/AbstractStatementValidator.java
@@ -23,19 +23,24 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 
 public abstract class AbstractStatementValidator<T extends AttestationStatement> implements AttestationStatementValidator {
-    @Override
-    public boolean supports(RegistrationObject registrationObject) {
-        AttestationStatement attestationStatement = registrationObject.getAttestationObject().getAttestationStatement();
+    private Class<?> parameterizedTypeClass;
 
+    public AbstractStatementValidator() {
         ParameterizedType parameterizedType = (ParameterizedType) getClass().getGenericSuperclass();
         Type actualTypeArgument = parameterizedType.getActualTypeArguments()[0];
 
         if (actualTypeArgument instanceof Class) {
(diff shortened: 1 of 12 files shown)


ynojima:
Interesting refactoring.

webauthn4j-core/src/main/java/com/webauthn4j/validator/attestation/statement/AbstractStatementValidator.java
```diff
@@ -0,0 +1,41 @@
...
+public abstract class AbstractStatementValidator<T extends AttestationStatement> implements AttestationStatementValidator {
+    @Override
+    public boolean supports(RegistrationObject registrationObject) {
+        AttestationStatement attestationStatement = registrationObject.getAttestationObject().getAttestationStatement();
+
+        ParameterizedType parameterizedType = (ParameterizedType) getClass().getGenericSuperclass();
+        Type actualTypeArgument = parameterizedType.getActualTypeArguments()[0];
+
+        if (actualTypeArgument instanceof Class) {
+            Class<?> aClass = (Class<?>) actualTypeArgument;
```
ynojima: `aClass` is better to be initialized in the constructor. No need to be calculated every time `supports` method called.

kvandermast:
Thanks, the idea was - initially - to use the generic when retrieving the attestation statement, so that the abstract class would have this

```
public <T> getAttestationStatement() {
   // ... 
}
```

I didn't manage to get to that part yet, but this seems like a potential refactoring candidate to reduce some implementation.

Putting it in the constructor seems a good approach indeed.

webauthn4j-core/src/main/java/com/webauthn4j/validator/attestation/statement/AbstractStatementValidator.java
```diff
@@ -0,0 +1,41 @@
...
+
+            return aClass.isAssignableFrom(attestationStatement.getClass());
+        }
+
+        return false;
```
ynojima: When does the execution reach here? I suppose it is better to throw `IllegalStateException` with message describing the situation.

kvandermast: In theory, never. 
- T must be of type AttestationStatement
- If the resolved type is not a class (covered by the above)

So an IllegalStateException might be better indeed.

PR #70, opened 2019-03-18
Proposed Signature implementation 

kvandermast:
An improvement suggestion on the signature implementation
- Limit the available signatures to the ones defined in `COSEAlgorithmIdentifier` and `JWAIdentifier`
- Couple the signer and verifier to the enum-value for ease of access.


webauthn4j-core/src/main/java/com/webauthn4j/signature/Signature.java
```diff
@@ -0,0 +1,92 @@
+package com.webauthn4j.signature;
+
+import com.webauthn4j.data.attestation.statement.COSEAlgorithmIdentifier;
+import com.webauthn4j.data.jws.JWAIdentifier;
+import com.webauthn4j.validator.exception.BadSignatureException;
+
+import java.security.*;
+
+public final class Signature {
```
ynojima: New class introduction brings complexity, and this class resolves little. I can't find needs for this change.

kvandermast: The idea is to align the supported algorithms and a the signature functionality, avoid the need to catch exceptions in case a unsupported algorithm is passed. The existence of the algorithms is already check when either the CoseAlgorithm or JWA id is verified in the creation/parsing of the objects.

ynojima: Where is room to pass unsupported algorithm? I suppose current implementation have already sealed it.

kvandermast: True, though the SignatureUtil allows to retrieve an instance (`createSignature`) of the java.signature.Signature class using a String as parameter name. IMHO, this suggestion provides a cleaner interface on signing and verifying the payload.

ynojima: SignatureUtil.createSignature takes a string as an argument and returns `java.signature.Signature`, but No public validator interface takes
`java.signature.Signature` as parameter.

PR #71, opened 2019-03-20
Moved SampleTrustAnchorProvider to tests

kvandermast:
`SampleTrustAnchorProvider` is not used in the lib itself, and is commented as "do not use in production", moved to test folder.


ynojima:
Thank you for suggestion!

PR #72, opened 2019-03-20
Added owasp dependency check task to scan for known CVEs

kvandermast:
Added owasp dependency check task to scan for known CVEs


ynojima:
Interesting plugin. Thank you for suggestion.

PR #73, opened 2019-03-21
Make HttpClient configurable

kvandermast:
A suggestion originating from our own server/network set-up. The current `SimpleHttpClient` would not be allowed/used in our set-up, nor would it be able to connect to anything remote.

My suggestion is to make the `HttpClient` used for interacting with the Fido MDS configurable:
- The `HttpClient` interface extends `AutoCloseable` to allow a `try-with-resouces` depending on the implementation of the HttpClient;
- An HttpClientFactory has been added to create an instance of the HttpClient, which will:
   - Look for a properties bundle `webauthn4j-metadata.properties` on the ClassLoaders classpath;
   - Check if there is a key available, name `webauthn4j.metadata.httpclient.impl`
   - Check if the class can be found on the classpath;
   - Check if the class implements the HttpClient interface
   - If any of the above fail, it will fall-back to the SimpleHttpClient implementation

The properties could look as such:
```properties
# Defines the class to use to create an outgoing connection for fetching the MDS data.
webauthn4j.metadata.httpclient.impl=com.webauthn4j.metadata.http.SimpleHttpClient
```


ynojima:
FidoMdsMetadataItemsProvider already have a constructor that allow caller inject `HttpClient`.
Why do you need a original mechanizm to lookup an implementation from property file?

ynojima:
For example, sample application of spring-security-webauthn, which adopts webauthn4j uses the constructor for custom `HttpClient` implementation injection.

https://github.com/sharplab/spring-security-webauthn/blob/master/samples/javaconfig/webauthn/fido-server-conformance-test/src/main/java/net/sharplab/springframework/security/webauthn/sample/app/config/WebSecurityBeanConfig.java#L154-L183

webauthn4j-metadata/src/main/java/com/webauthn4j/metadata/http/HttpClient.java
```diff
@@ -14,13 +14,13 @@
...
 
 /**
  * HTTP Client for FIDO FidoMdsMetadataItemImpl Service
  */
-public interface HttpClient {
-
-    String fetch(String uri);
-
+public interface HttpClient extends AutoCloseable {
+    String fetch(String uri) throws MDSException;
```
ynojima: Adding `throws MDSException` is reasonable. Thanks.

kvandermast:
For some reason I overlooked that one entirely, as I assumed that this would be invoked within the Fido statement validator. Must have mixed that one up with the Spring example. 
We make it configurable because depending on our environments, we need different implementations and security constraints... Seems I was too much focused on the actual usage of the SimpleHttpClient(), which would break on our environments.

webauthn4j-metadata/src/main/java/com/webauthn4j/metadata/http/SimpleHttpClient.java
```diff
@@ -26,12 +28,18 @@
...
+    public SimpleHttpClient() {
+    }
 
     @Override
-    public String fetch(String url) {
+    public String fetch(String url) throws MDSException {
         try {
             URL fetchUrl = new URL(url);
-            HttpURLConnection urlConnection = (HttpURLConnection) fetchUrl.openConnection();
+            urlConnection = (HttpURLConnection) fetchUrl.openConnection();
```
ynojima: If `fetch` method is invoked in parallel, this may cause trouble.

ynojima:
To keep the implementation simple, and there is way to inject `HttpClient`, I close this issue.

PR #84, opened 2019-04-30
Static secure random

kvandermast:
We've run into a problem that the seeding on creating new SecureRandoms takes too much time (taking up more than 20 - 25s/req). We foresee a large load, so my suggestion would be to make this static/factory based. What do you think?

### DIFF
--- a/webauthn4j-test/src/main/java/com/webauthn4j/test/authenticator/u2f/FIDOU2FAuthenticator.java
+++ b/webauthn4j-test/src/main/java/com/webauthn4j/test/authenticator/u2f/FIDOU2FAuthenticator.java
@@ -60,9 +60,7 @@ public class FIDOU2FAuthenticator {
         byte[] challengeParameter = registrationRequest.getChallengeParameter();
         byte[] applicationParameter = registrationRequest.getApplicationParameter();
 
-        SecureRandom secureRandom = new SecureRandom();
-        byte[] nonce = new byte[32];
-        secureRandom.nextBytes(nonce);
+        byte[] nonce = SecureRandomFactory.getSecureRandom().randomBytes(32);
         KeyPair keyPair = getKeyPair(applicationParameter, nonce);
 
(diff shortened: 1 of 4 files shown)


ynojima:
Thank you for pointing out. That need to be improved. will look into your pull-request.

ynojima:
`SecureRandomFactory` sounds like factory class for `SecureRandom`, but it isn't. It's confusing.

BTW, since SecureRandom is only used in test code or test utilities, it won't be overhead for request processing.

ynojima:
As I suppose having `SecureRandom` per classes won't make large overhead, I made PR without factory.
https://github.com/webauthn4j/webauthn4j/pull/85

kvandermast:
As long as it is only created once, it will not be a large overhead. 

I first called the `SecureRandomFactory` a `SecureRandomHelper`, then a `SecureRandomUtil`... I wasn't convinced by the naming either. Have a SecureRandomUtil/-Helper class which is publicly available would be a plus too, as you can suggest users, for generating a challenge for the options, to use that class. Avoiding overhead on that side again.

It's not the first time that I notice that I change code that is solely used for test purposes (e.g. being in the util module). Isn't there a way to regroup the classes so that, if they are only used for unit testing, are moved to the `test` module?

ynojima:
> Have a SecureRandomUtil/-Helper class which is publicly available would be a plus too, as you can suggest users, for generating a challenge for the options, to use that class. Avoiding overhead on that side again.

Hmm, good point. I recalled that `SecureRandom` is used inside `DefaultChallenge`. `DefaultChallenge` uses `UUID` to generate a random byte array, and `UUID` uses `SecureRandom` and cache it in the static field.

> It's not the first time that I notice that I change code that is solely used for test purposes (e.g. being in the util module). Isn't there a way to regroup the classes so that, if they are only used for unit testing, are moved to the test module?

Sorry for confusion. `KeyUtil` class is better to be moved to `test` module. will do.

PR #85, opened 2019-05-01
Improve SecureRandom handling

ynojima:
* Stop using `SecureRandom.getInstanceStrong()` as it makes large overhead, and I don't see value on using it
* Have `SecureRandom` per classes to reduce instance creation overhead.

### DIFF
--- a/webauthn4j-util/src/main/java/com/webauthn4j/util/KeyUtil.java
+++ b/webauthn4j-util/src/main/java/com/webauthn4j/util/KeyUtil.java
@@ -29,6 +29,8 @@ import java.security.spec.RSAKeyGenParameterSpec;
  */
 public class KeyUtil {
 
+    private static SecureRandom secureRandom = new SecureRandom();
+
     private KeyUtil() {
     }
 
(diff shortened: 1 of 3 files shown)


webauthn4j-util/src/main/java/com/webauthn4j/util/KeyUtil.java
```diff
@@ -53,13 +55,13 @@ public static KeyPairGenerator createECKeyPairGenerator() {
...
     public static KeyPair createECKeyPair(byte[] seed, ECParameterSpec ecParameterSpec) {
         KeyPairGenerator keyPairGenerator = createECKeyPairGenerator();
-        SecureRandom random = null;
+        SecureRandom random;
         try {
             if (seed != null) {
                 random = SecureRandom.getInstance("SHA1PRNG"); // to make it deterministic
                 random.setSeed(seed);
             } else {
-                random = SecureRandom.getInstanceStrong();
```
ynojima: Since `SecureRandom.getInstanceStrong()` returns `SecureRandom` with `NativePRNGBlocking`, it makes large overhead even when `SecureRandom.nextBytes()` is called.
I don't see value on using `SecureRandom.getInstanceStrong()` here.

webauthn4j-util/src/main/java/com/webauthn4j/util/KeyUtil.java
```diff
@@ -29,6 +29,8 @@
  */
 public class KeyUtil {
 
+    private static SecureRandom secureRandom = new SecureRandom();
```
kvandermast: make this final

ynojima: Overlooked your comment.
Will be fixed in #86

PR #87, opened 2019-05-16
FIX json mapping

hc42:
JsonTypeInfo duplicates existing propery "1" of subclasses resulting in
invalid json and failure on deserializion

before serialization result looked like {"1":"2","2":null,"3":-7," ..... ,"1":2} with duplicate keys
after fix like {"2":null,"3":-7," ..... ,"1":2} and deserializion works

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/data/attestation/statement/AttestationStatement.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/data/attestation/statement/AttestationStatement.java
@@ -16,6 +16,7 @@
 
 package com.webauthn4j.data.attestation.statement;
 
+import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
 import java.io.Serializable;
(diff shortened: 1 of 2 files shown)


webauthn4j-core/src/main/java/com/webauthn4j/data/attestation/statement/AttestationStatement.java
```diff
@@ -24,7 +25,16 @@
  * Attestation metadata.certs container
  */
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "format")
+@JsonSubTypes({
```
ynojima: Subtypes are defined here:   https://github.com/webauthn4j/webauthn4j/blob/f92dee2674ebbe2c5b9b9b91fcac0f3bc266ae85/webauthn4j-core/src/main/java/com/webauthn4j/converter/jackson/WebAuthnCBORModule.java#L63

hc42:
Thanks for your quick feedback. Indeed using CborConverter this works in both cases. However JSON will not work (also  using CborConverter->getJsonConverter) will produce invalid JSON (without these annotations). For me CBOR is ok, however using the annotations might be more general / convenient.

ynojima:
I don't recommend you to serialize attestation statements to JSON as the attestation statement serializers are not designed to serialize to JSON.

PR #101, opened 2019-06-16
CborMapper and JsonMapper threadsafety

kvandermast:
Both the `CborMapper` and `JsonMapper` are not thread safe on initialization of the `WebAuthnCborModule` and `WebAuthnJSONModule`. 

```
   /**
     * Returns the {@link ObjectMapper} configured for CBOR processing
     *
     * @return the {@link ObjectMapper} configured for CBOR processing
     */
    private ObjectMapper getCborMapper() {
        // not thread safe...
        if (!cborMapperInitialized) {
            cborMapper.registerModule(new WebAuthnCBORModule(getJsonConverter(), this));
            cborMapper.configure(DeserializationFeature.WRAP_EXCEPTIONS, false);
            cborMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
            cborMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
            cborMapperInitialized = true;
        }
        return cborMapper;
    }

    /**
     * Returns the twined {@link JsonConverter}
     *
     * @return the twined {@link JsonConverter}
     */
    public JsonConverter getJsonConverter() {
        if (jsonConverter == null) {
            jsonConverter = new JsonConverter(jsonMapper, cborMapper);
        }
        return jsonConverter;
    }
```
Jackson's ObjectMappers are by design thread safe:

```
com.fasterxml.jackson.databind.ObjectMapper


*<p> 
* Mapper instances are fully thread-safe provided that ALL configuration of the
* instance occurs before ANY read or write calls. If configuration of a mapper instance
* is modified after first usage, changes may or may not take effect, and configuration
* calls themselves may fail.
* If you need to use different configuration, you have two main possibilities:
*<ul>
* <li>Construct and use {@link ObjectReader} for reading, {@link ObjectWriter} for writing.
* Both types are fully immutable and you can freely create new instances with different
* configuration using either factory methods of {@link ObjectMapper}, or readers/writers
* themselves. Construction of new {@link ObjectReader}s and {@link ObjectWriter}s is
* a very light-weight operation so it is usually appropriate to create these on per-call
* basis, as needed, for configuring things like optional indentation of JSON.
* </li>
* <li>If the specific kind of configurability is not available via {@link ObjectReader} and
* {@link ObjectWriter}, you may need to use multiple {@link ObjectMapper} instead (for example:
* you cannot change mix-in annotations on-the-fly; or, set of custom (de)serializers).
* To help with this usage, you may want to use method {@link #copy()} which creates a clone
* of the mapper with specific configuration, and allows configuration of the copied instance
* before it gets used. Note that {@link #copy} operation is as expensive as constructing
* a new {@link ObjectMapper} instance: if possible, you should still pool and reuse mappers
* if you intend to use them for multiple operations.
* </li>
* </ul>
*<p>
```

As indicated by the comment, the ObjectMappers are recreated with every validation or registration. Thus they could benefit from being implemented as a single instance. Applying this also removes the need for the cyclic dependency of passing either the CborMapper or the JsonMapper to their corresponding module.

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/converter/util/JsonConverter.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/converter/util/JsonConverter.java
@@ -24,10 +24,8 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.exc.MismatchedInputException;
-import com.fasterxml.jackson.dataformat.cbor.CBORFactory;
 import com.webauthn4j.converter.exception.DataConversionException;
 import com.webauthn4j.converter.jackson.WebAuthnJSONModule;
-import com.webauthn4j.util.AssertUtil;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -39,31 +37,26 @@ import java.io.UncheckedIOException;
  */
 public class JsonConverter implements Serializable {
 
+    public final static JsonConverter INSTANCE = new JsonConverter();
     private static final String INPUT_MISMATCH_ERROR_MESSAGE = "Input data does not match expected form";
 
     /**
      * As it may not be initialized, jsonMapper must be used through getJsonMapper method
      */
     private ObjectMapper jsonMapper;
-    private ObjectMapper cborMapper;
 
-    private boolean jsonMapperInitialized = false;
 
-    private CborConverter cborConverter;
-
-    public JsonConverter(ObjectMapper jsonMapper, ObjectMapper cborMapper) {
-        AssertUtil.notNull(jsonMapper, "jsonMapper must not be null");
-        AssertUtil.notNull(cborMapper, "cborMapper must not be null");
-
-        AssertUtil.isTrue(!(jsonMapper.getFactory() instanceof CBORFactory), "factory of jsonMapper must be JsonFactory.");
-        AssertUtil.isTrue(cborMapper.getFactory() instanceof CBORFactory, "factory of cborMapper must be CBORFactory.");
-
-        this.jsonMapper = jsonMapper;
-        this.cborMapper = cborMapper;
+    private JsonConverter() {
+        this(new ObjectMapper(new JsonFactory()));
     }
 
-    public JsonConverter() {
-        this(new ObjectMapper(new JsonFactory()), new ObjectMapper(new CBORFactory()));
+    public JsonConverter(ObjectMapper jsonMapper) {
+        this.jsonMapper = jsonMapper;
+
+        jsonMapper.registerModule(new WebAuthnJSONModule());
+        jsonMapper.configure(DeserializationFeature.WRAP_EXCEPTIONS, false);
+        jsonMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        jsonMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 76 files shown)


ynojima:
> Both the CborMapper and JsonMapper are not thread safe

Thank you for pointing out. It need to be fixed.

> As indicated by the comment, the ObjectMappers are recreated with every validation or registration.

From which comment did you think so?
A WebAuthnAuthenticationContextValidator instance reuses a same JsonConverter instance all over the validator life time.

webauthn4j-core/src/main/java/com/webauthn4j/converter/util/JsonConverter.java
```diff
@@ -39,31 +37,26 @@
  */
 public class JsonConverter implements Serializable {
 
+    public final static JsonConverter INSTANCE = new JsonConverter();
```
ynojima: I don't take singleton pattern for JsonConverter/CborConverter. In some cases, a user of WebAuthn4J would like to register his own class to ObjectMapper owned by one WebAuthnAuthenticationContextValidator instance, but he may not want to register the class to the other WebAuthnAuthenticationContextValidator instance.

kvandermast:
> A WebAuthnAuthenticationContextValidator instance reuses a same JsonConverter instance all over the validator life time.

Well yes, that's just the idea: every time you create a WAACV, you create a new JsonConverter and CborConvertor, thus reregistering the Modules for serialization. In our case, we treat about 4.5 million authentications a day, so we don't want to recreate the Converters 4.5 million times. 

I also don't understand the reasoning behind a user wanting to register his own ObjectMapper, as it would be shielded anyhow from his implementation? Would it the not be better to have the validators implement and abstract class or interface, so that they can manage this themselves? 

PS, thinking of which, why do the Converters implement `Serializable`? As far as I can see, they are not serialized themselves? (Just asking, no blame)

ynojima:
> Well yes, that's just the idea: every time you create a WAACV, you create a new JsonConverter and CborConvertor, thus reregistering the Modules for serialization. 

Why don't you reuse WAACV?

> In our case, we treat about 4.5 million authentications a day, so we don't want to recreate the Converters 4.5 million times.

That's huge! Please let me know when you successfully introduced webauthn4j into production.

> Would it the not be better to have the Cbor/JsonConverters implement and abstract class or interface, so that they can manage this themselves?

Indroducing interface for Cbor/JsonConverters seems good idea. I'll check its feasibility.

AuthenticatorDataSerializer, which holds CborConverter in its field, extends StdSerializer, which implements Serializable.
That's why I made it Serializable, but maybe it's better to make the CborConverter field transient.

ynojima:
If you don't want to reuse WAACV, reuse JsonConverter by injecting the same converter instance through the constructor of WAACV.

kvandermast:
> AuthenticatorDataSerializer, which holds CborConverter in its field, extends StdSerializer, which implements Serializable.
> That's why I made it Serializable, but maybe it's better to make the CborConverter field transient.

Ok, could it maybe also benefit from optimizing the creation of the AuthenticatorDataConverter?, e.g. 

```java
/**
 * Jackson Serializer for {@link AuthenticatorData}
 */
public class AuthenticatorDataSerializer extends StdSerializer<AuthenticatorData> {

    private static final AuthenticatorDataConverter AUTHENTICATOR_DATA_CONVERTER = new AuthenticatorDataConverter(CborConverter.INSTANCE);

    public AuthenticatorDataSerializer() {
        super(AuthenticatorData.class);
    }

    /**
     * {@inheritDoc}
     */
    @Override
    public void serialize(AuthenticatorData value, JsonGenerator gen, SerializerProvider provider) throws IOException {
        gen.writeBinary(AUTHENTICATOR_DATA_CONVERTER.convert(value));
    }
}
```

kvandermast:
> If you don't want to reuse WAACV, reuse JsonConverter by injecting the same converter instance through the constructor of WAACV.

That would be doable, but for that I would suggest that both the Json and CborConverter implement a interface defining the read-/write-operations. If you want, I can have a look at that. 
This way, the library we wrapped around yours, can implement a singleton for both and use/re-use that one on the RP server.

ynojima:
> Indroducing interface for Cbor/JsonConverters seems good idea. I'll check its feasibility.

I thought introducing interface may make room for users to use library other than Jackson, but I cannot foresee sufficient amount of needs.

ynojima:
Fixed by #105

PR #103, opened 2019-06-20
Jackson thread safety and cyclic dependencies (rev.)

kvandermast:
I've added a new branch which should resolve the thread safety issues, and also does not use singletons on the JsonConverter and CborConverter.

Let me know what you think of this approach.

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/converter/AuthenticatorDataConverter.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/converter/AuthenticatorDataConverter.java
@@ -45,10 +45,12 @@ public class AuthenticatorDataConverter extends AbstractCborConverter {
     // ================================================================================================
 
     public AuthenticatorDataConverter() {
-        super(new ObjectMapper(new CBORFactory()),
-                new WebAuthnAuthenticatorModule(),
-                false,
-                false);
+        this(new ObjectMapper(new CBORFactory()));
+    }
+
+    public AuthenticatorDataConverter(ObjectMapper mapper) {
+        super(mapper, new WebAuthnAuthenticatorModule());
+
     }
 
     //~ Methods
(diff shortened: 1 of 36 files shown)


ynojima:
"WebAuthn4J user can add his/her own webauthn extension support."
That's what I'd like to achieve while designing JsonConver/CborConver/serializers/deserializers.
To realize it, WebAuthn4J needs to allow custom serializer / deserializer registration to its internal Jackson ObjectMapper.
On the other hand, that customization must be kept in a particular WebAuthnRegistrationContextValidator / WebAuthnAuthenticationContextValidator instance and its callee validators. 
It must be avoided that customization to an instance affects other instances.

That's why I pass JsonConverter and CborConverter from caller validators to callee validators.

I won't adopt singleton pattern like this:
https://github.com/webauthn4j/webauthn4j/blob/d0cf244d363a55e23921e74ab1b1f57942bd732b/webauthn4j-core/src/main/java/com/webauthn4j/converter/jackson/deserializer/AuthenticatorDataDeserializer.java#L43

If you would like to avoid ObjectMapper initialization overhead, reuse WARCV/WAACV instance, or inject a specific ObjectMapper instance via WARCV/WAACV constructor.

kvandermast:
> If you would like to avoid ObjectMapper initialization overhead, reuse WARCV/WAACV instance, or inject a specific ObjectMapper instance via WARCV/WAACV constructor.

I follow you on the above, that why this one does not provide a singleton for the Cbor-/JsonConverter. I've added in our own library a thread safe creation of these classes as we use them in our own code too. The largest problem is the cyclic dependency: when you initialize the module for Jackson, you pass the converter as a parameter (which might not have been initialized too btw). 

The Cbor and JsonConverter also instantiate instance of each other (and are using them), which seems a bad design to me in the light of atomic responsibilities.

ynojima:
> The largest problem is the cyclic dependency: when you initialize the module for Jackson, you pass the converter as a parameter (which might not have been initialized too btw).
> The Cbor and JsonConverter also instantiate instance of each other (and are using them), which seems a bad design to me in the light of atomic responsibilities.

I agree, I'd like to solve it in a clean way, but I couldn't. It's inevitable as JsonConverter needs CborConverter, and CborConverter needs JsonConverter.
I made a pull-request to ensure thread safety.
https://github.com/webauthn4j/webauthn4j/pull/104

ynojima:
Fixed by #105

PR #104, opened 2019-06-20
Ensure thread safety on JsonConverter/CborConverter

ynojima:


webauthn4j-core/src/main/java/com/webauthn4j/converter/util/CborConverter.java
```diff
@@ -134,11 +137,15 @@ public JsonNode readTree(byte[] bytes) {
      */
     private ObjectMapper getCborMapper() {
         if (!cborMapperInitialized) {
-            cborMapper.registerModule(new WebAuthnCBORModule(getJsonConverter(), this));
-            cborMapper.configure(DeserializationFeature.WRAP_EXCEPTIONS, false);
-            cborMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-            cborMapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-            cborMapperInitialized = true;
+            synchronized (cborMapperInitializationLock){
```
kvandermast: Less performant because it locks with every call to getCborMapper()

ynojima: After cborMapper is initialized, it won't reach here because of the `if` statement 1 line above.

kvandermast: Yes, but in a multi threaded env you might a couple of pending threads locked at the same time still, as they will pass the first check of non-initialization. This is a work-around because the refactoring to the initialization on construction of the converter would inevitably lead to StackOverflows because of the mutual dependencies.

ynojima: > you might a couple of pending threads locked at the same time still, as they will pass the first check of non-initialization.

At first, yes, but eventually resoloved.

> This is a work-around because the refactoring to the initialization on construction of the converter would inevitably lead to StackOverflows because of the mutual dependencies.

Yes. but the mutual dependencies cannot be split off. These classes are twin. They need each other.
How do you think about the approach#2 (https://github.com/webauthn4j/webauthn4j/pull/105)? I removed locks.

ynojima:
Fixed by #105

Issue #160, opened 2019-10-07
Verifying an UserHandle


kasecato:
**Is your feature request related to a problem? Please describe.**

In the verifying assertion flow, you should to verify `response.userHandle`.

**Describe the solution you'd like**

> 2. Identify the user being authenticated and verify that this user is the owner of the public key credential source credentialSource identified by credential.id:

>> If the user was identified before the authentication ceremony was initiated,
>>> verify that the identified user is the owner of credentialSource. If credential.response.userHandle is present, verify that this value identifies the same user as was previously identified.

>> If the user was not identified before the authentication ceremony was initiated,
>>> verify that credential.response.userHandle is present, and that the user identified by this value is the owner of credentialSource.

https://www.w3.org/TR/webauthn-1/#verifying-assertion

**Describe alternatives you've considered**

None.

**Additional context**

None.

ynojima:
After closer investigation, this issue is classified out of WebAuthn4J's scope. user identification is caller application's responsibility.

https://github.com/webauthn4j/webauthn4j/blob/master/webauthn4j-core/src/main/java/com/webauthn4j/validator/WebAuthnAuthenticationContextValidator.java#L101-L120

PR #170, opened 2019-10-10
Add userHandle to WebAuthnAuthenticationContext

ynojima:
Cont. to #171 

To work on #160, WebAuthnAuthenticationContext needs to have userHandle menber


ynojima:
> but not treat it by webauthn4j because the handling of User Handle depends on the caller application's context strongly.

No. I'll implement userHandle verification logic later in a separate pull-request.

PR #231, opened 2020-02-24
Fix documentation - use credentialId from authenticationData in authentication

bedrin:
Fix documentation - use credentialId from authenticationData instead of attestedCredentialData in authentication flow

As per specification "For assertion signatures, the AT flag MUST NOT be set and the attestedCredentialData MUST NOT be included."
https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ catch (ValidationException e){
 }
 // please update the counter of the authenticator record
 updateCounter(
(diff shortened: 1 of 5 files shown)


ynojima:
Good catch. Thank you for pointing out!

PR #251, opened 2020-04-02
Remove unnecessary gradle config for clean up

ynojima:


ynojima:
It is necessary, and fixed in https://github.com/webauthn4j/webauthn4j/pull/264

Issue #260, opened 2020-04-15
Webauthn4j crashes when setting requireResidentKey to true in Chrome


arianvp:
**Describe the bug**

**To Reproduce**
1. In for example keycloak,  mandate requireResidentKey, 
2. Try to register a credential
3. Serialisation fails: with following stacktrace

```

14:32:22,586 DEBUG [org.keycloak.authentication.requiredactions.WebAuthnRegister] (default task-16) Input data does not match expected form: com.webauthn4j.converter.exception.DataConversionException: Input data does not match expected form
	at com.webauthn4j.webauthn4j-core//com.webauthn4j.converter.util.CborConverter.readValue(CborConverter.java:99)
	at com.webauthn4j.webauthn4j-core//com.webauthn4j.converter.AuthenticatorDataConverter.convertToExtensions(AuthenticatorDataConverter.java:154)
	at com.webauthn4j.webauthn4j-core//com.webauthn4j.converter.AuthenticatorDataConverter.convert(AuthenticatorDataConverter.java:131)
	at com.webauthn4j.webauthn4j-core//com.webauthn4j.converter.jackson.deserializer.AuthenticatorDataDeserializer.deserialize(AuthenticatorDataDeserializer.java:63)
	at com.webauthn4j.webauthn4j-core//com.webauthn4j.converter.jackson.deserializer.AuthenticatorDataDeserializer.deserialize(AuthenticatorDataDeserializer.java:33)
	at com.fasterxml.jackson.core.jackson-databind@2.9.10.1//com.fasterxml.jackson.databind.deser.SettableBeanProperty.deserialize(SettableBeanProperty.java:530)
	at com.fasterxml.jackson.core.jackson-databind@2.9.10.1//com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeWithErrorWrapping(BeanDeserializer.java:528)
	at com.fasterxml.jackson.core.jackson-databind@2.9.10.1//com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeUsingPropertyBasedWithExternalTypeId(BeanDeserializer.java:945)
	at com.fasterxml.jackson.core.jackson-databind@2.9.10.1//com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeWithExternalTypeId(BeanDeserializer.java:853)
[...]

PR #261, opened 2020-04-17
Align unknown extension handling to WebAuthn Level2 draft1

ynojima:
When an unknown extension is included in the serialized data, `DataConversionException` was thrown (https://github.com/webauthn4j/webauthn4j/blob/0.11.0.RELEASE/webauthn4j-core/src/test/java/com/webauthn4j/converter/AuthenticationExtensionsClientInputsConverterTest.java#L64-L68) while deserializing.

In W3C WebAuthn spec level1, it was requested to reject unintended extension (https://www.w3.org/TR/webauthn/#registering-a-new-credential), but in level2 spec, rejecting unintended extension is left to relying party decision (https://www.w3.org/TR/2019/WD-webauthn-2-20191126/#sctn-registering-a-new-credential).

Since rejecting unknown extension is not welcomed ( https://github.com/webauthn4j/webauthn4j/issues/260 ), this PR changes the behavior.
Now `DataConversionException` won't be thrown in the case, and `UnknownExtensionAuthenticatorOutput` | `UnknownExtensionClientInput` | `UnknownExtensionClientOutput` class is included in the deserialized data.


ynojima:
As the SonarCloud warns about code duplication, it can be ignored.

PR #296, opened 2020-05-30
Add uvm extension

ynojima:
Moved UserVerificationMethod class to core module
Moved KeyProtection class to core module and renamed to KeyProtectionType
Moved MathcerProtection class to core module and renamed to MatcherProtectionType


ynojima:
This is addressed in #306. Closing.

PR #300, opened 2020-06-07
[WIP] Add shiftleft pipeline for evaluation

ynojima:


ynojima:
With this evaluation, it turned out that ShiftLeft doesn't expose results to the public (only available to registered users). As it can be a hurdle for new contributors, we don't adopt it for now.

PR #326, opened 2020-08-17
Support for Apple App Attest (apple-appattest)

veehaitch:
This PR introduces support for the Apple App Attest attestation format (`apple-appattest`) available with the release of iOS 14 (expected for the end of September 2020). For details, please refer to the official articles [on Apple Developer](https://developer.apple.com/documentation/devicecheck/dcappattestservice).

Please consider this a draft. I just wanted to publish this early on to gather your feedback and to avoid double work.

For a full validation of the attestation (including the app ID and key ID, but excluding the receipt) also see https://github.com/veehaitch/devicecheck-appattest


ynojima:
Thank you for your pull-request! I'll look into.

ynojima:
In my understanding, Apple App Attest is an attestation format for iOS app, not for WebAuthn.
`DCAppAttestService` does not generate WebAuthn [PublicKeyCredential](https://w3c.github.io/webauthn/#iface-pkcredential), but an attestation statement only. 
Since current WebAuthn4J is designed to validate WebAuthn PublicKeyCredential, its API signature is not useful for users who want to validate Apple App Attest if we only add a parser and validator for Apple App Attest.
I found you add `RegistrationObject` test data for Apple App Attest, which includes `CollectedClientData`, but is it generated by Apple device? or hand crafted only to fit WebAuthn4J API signature?


I suppose the entry point API for validating Apple App Attest should be dedicated from Validator for WebAuthn. How do you think? Please correct me if I am wrong.

veehaitch:
Thanks for your response! According to the available iOS 14 APIs and the documentation, I’d suppose you’re right:

> The App Attest service creates an attestation object composed of authenticator data and an attestation statement, as specified by the [Web Authentication](https://www.w3.org/TR/webauthn/#sec-authenticator-data) specification

Yes, I’ve created the registration object on my own using the iOS 14 Beta.

I’m not very familiar with the WebAuth4J codebase and API’s; hence, I don’t know how or—even if—this fits in best. I was hoping that you'd know 🙂  I just thought that this project might be the right place to make the new feature available to many people in a timely fashion. Of course, I might be mistaken.

As far as I understand, however, Android Key Attestation—being explicitly supported by this project—works very similar to the new iOS App Attest Service. How do they differ? After all, the iOS APIs seem to have been created with at least WebAuthn in mind.

ynojima:
> Yes, I’ve created the registration object on my own using the iOS 14 Beta.

I know Attestation object is from your iOS 14 Beta, but I assume [the CollectedClientData you provided](https://github.com/webauthn4j/webauthn4j/blob/b8897384419d20b285a7bd3fb7514e9b71687290/webauthn4j-test/src/main/java/com/webauthn4j/test/TestDataUtil.java#L119) is not from iOS but you created as WebAuthn4J requires it. Is my assumption wrong?

I mean I suppose preparing WebAuthn-style CollectedClientData is not welcomed, and unnecessary requirement for library users who just want to validate iOS app runtime consistency.
I'll make a PoC of Apple App Attest validator in this weekend based on your artifact.


> As far as I understand, however, Android Key Attestation—being explicitly supported by this project—works very similar to the new iOS App Attest Service. How do they differ? After all, the iOS APIs seem to have been created with at least WebAuthn in mind.

I suppose Android SafetyNet Attestation, which is also supported by WebAuthn4J, is more similar to Apple App Attest.
Both Android SafetyNet API and Apple App Attest assure an App is not running on compromised environment. 
Android SafetyNet is also used for WebAuthn, that's why WebAuthn4J supports Android SafetyNet Attestation validation.
But Apple App Attest will not be used for WebAuthn because Apple published another dedicated attestation format for WebAuthn, which is named "Apple Anonymous Attestation". (WebAuthn4J will support it. [Here is a PoC for it](https://github.com/webauthn4j/webauthn4j-spring-security/pull/399))
Apple App Attest's format is similar to WebAuthn, but will not be used for WebAuthn. This is the difference I see.

Apple App Attest and WebAuthn shares same technology, similar format and I agree WebAuthn4J is a good place to implement validator for it, but Its API signature will be differ because of its usage.
Anyway, I'll prepare a PoC for it. Please give me a few days :)

veehaitch:
> I know Attestation object is from your iOS 14 Beta, but I assume the CollectedClientData you provided is not from iOS but you created as WebAuthn4J requires it. Is my assumption wrong?

No, you're right! Sorry, if I didn't express myself very well.

> I suppose Android SafetyNet Attestation, which is also supported by WebAuthn4J, is more similar to Apple App Attest.
> Both Android SafetyNet API and Apple App Attest assure an App is not running on compromised environment.

I'm not sure I'd agree here. I'd actually say that Apple App Attest combines all the core features of Android Key Attestation _and_ Safety Net in the same call.

First and foremost, Apple App Attest enables key attestation. Apple has been supporting the creation of keys in secure hardware (the Secure Enclave, SE) for quite a long time already. The SE ensures that the private key always remains within the trusted execution environment and never enters user space. Until iOS 14, however, it lacked any attestation: all you'd send to your backend has been a P-256 public key (by the way, the SE only supports P-256) without any trustworthy statement and, hence, no option for the receiving party to verify that the corresponding private key was in fact created in the SE.

Apple's approach is in fact very similar to Android's key attestation: both require a challenge for the creation of the key and then provide you with a certificate chain as a result: the leaf certificate contains the public key and an X.509 extension which contains the challenge for verification (Android provides quite a few further properties regarding the key capabilities, which Apple does not have). To verify the authenticity, one has to verify the certificate chain up to a Google or Apple root CA.

The attestation statement does, however, also contain properties (within the `authenticatorData`) which Android Key Attestation itself does not provide but only Safety Net. This applies, for example, to the identifier of the app which created the key and attestation statement. On Android, you can only retrieve this information through an _additional_ call to Safety Net (NB: this makes it inferior to App Attest) which also gives you additional information on the integrity of the environment. Apple App Attest won't, as far as I understand, create a key and attestation attestation if it considers the environment compromised. Both, Safety Net and App Attest involve an online call to Google or Apple, respectively.

I'm mostly interested in using attestation to securely enroll a public key created within an iOS or Android app with a backend service. Afterward, the backend could, for example, require the app to sign all HTTP requests with the attested public key to ensure the request originated from a trusted device.

As Android Safety Net and Android Key Attestation are part of the official Web Authentication standard and Apple seemingly joins this initiative with App Attest, I thought it makes sense to leverage an approach which involves procedures from WebAuthn—or at least parts of it. I also agree with you that forcing the creation of `CollectedClientData` is not the best approach. For an additional API of this library, it would be useful to also verify additional app properties as part of the attestation validation. Cf. [`AttestationValidator`](https://github.com/veehaitch/devicecheck-appattest/blob/master/src/main/kotlin/ch/veehait/devicecheck/appattest/AttestationValidator.kt#L44-L49) of my PoC. The same applies to the additional properties of Android Key Attestation and Safety Net.

> But Apple App Attest will not be used for WebAuthn because Apple published another dedicated attestation format for WebAuthn, which is named "Apple Anonymous Attestation". [...] Apple App Attest's format is similar to WebAuthn, but will not be used for WebAuthn.

Are you sure? I mean, it appears useful to me to allow an app to talk to a WebAuthn endpoint which requires App Attest. I don't see why this should be limited to Safari / a webview.

> Anyway, I'll prepare a PoC for it. Please give me a few days :)

Of course, take your time. Thanks for the quick response again. Also, I'm happy to lend a hand! Just let me now :)

ynojima:
You are right, Apple App Attest also enables key attestation. From that point of view, Apple App Attest does the similar role of Android Key Attestation.

I made a PoC for Apple App Attest validation based on your pull-request. It still needs polishment, but describes direction.
https://github.com/webauthn4j/webauthn4j/pull/329

PoC test code:
https://github.com/webauthn4j/webauthn4j/blob/a01cc2b380132da4574f17f5e2941c7f777b8491/webauthn4j-device-check/src/test/java/com/webauthn4j/appattest/DeviceCheckAttestationManagerTest.java#L36-L51

Core logic of WebAuthn validation are segregated into new "Core" prefixed classes, and WebAuthn specific members like collectedClientData and clientDataJSON are no longer required.

ynojima:
Now your commits are merged into master with some modifications. Thanks!

PR #328, opened 2020-08-18
Relax origin scheme requirement within Origin class

ynojima:
Relax origin scheme requirement within Origin class to accept "android" scheme from Android FIDO2 API.
Even if illegal scheme is provided, `OriginValidator` will reject the request by comparing with valid data.

`Origin#getPort` now returns `Integer` instread of `int` because non `https` or `http` schemes may not have `port`.


ynojima:
This pull-request is superseded by #354

PR #330, opened 2020-08-24
Deprecate expectedExtensionIds member

ynojima:
from Parameters classes

As matching extension id(s) and expected list is no longer required from
WebAuthn Level 2 spec, deprecate the feature.
The feature will be removed by WebAuthn4J 1.0 GA release.


ynojima:
`@Deprecated` annotations make SonarCloud check fail, but the deprecation is temporary and can be ignored. Merging.

PR #350, opened 2020-09-26
Add doc about how to serialize/deserialize authenticator

ynojima:


docs/src/reference/asciidoc/en/deep-dive.adoc
```diff
@@ -1,5 +1,108 @@
...
+If you would like to persist as `String`, use `Base64UrlUtil` to convert from `byte[]` to base64url `String`.
+
+[source,java]
+----
+AttestedCredentialDataConverter attestedCredentialDataConverter = new AttestedCredentialDataConverter(objectConverter);
+
+// serialize
+byte[] serialized = attestedCredentialDataConverter.convert(attestedCredentialData);
+// deserialize
+AttestedCredentialDataConverter deserialized = attestedCredentialDataConverter.convert(serialized);
```
daniel-mezo, line 31: Hi @ynojima, there is a typo here, the static type should be ```AttestedCredentialData``` I believe.

daniel-mezo, line 31: Hi @ynojima, there is a typo here, the static type should be ```AttestedCredentialData```.

ynojima: Thank you for pointing out! 
Fixed in https://github.com/webauthn4j/webauthn4j/pull/603

Issue #358, opened 2020-10-07
"ArrayIndexOutOfBoundsException: 1" on X.509 name parsing


staranchenko:
[...]

PR #359, opened 2020-10-08
Bugfix: Failed to parse attestatation certificate DN

ynojima:
Addresses #358

Breaking change:

- `X500Name` and `Attributes` classes are removed, but these are designed for internal use, I suppose no one is affected.


ynojima:
Duplication is by design. Merging.

PR #372, opened 2020-11-02
Clarify nullability to better Kotlin interoperability

ynojima:
- Add `@NonNull` and `@Nullabe` annotations to all the classes in the webauthn4j-core and webauthn4j-util module


ynojima:
SonarCloud warnings will be resolved in the separate pull-requests.

PR #390, opened 2020-11-05
Reformat code

ynojima:


ynojima:
duplication is by design. merging.

PR #409, opened 2020-11-23
A relaxed origin validator

onne:
The spec mandates a relaxed same origin restriction, as opposed to an equality test - see https://www.w3.org/TR/webauthn/#determines-the-set-of-origins-on-which-the-public-key-credential-may-be-exercised.

This change includes:
- A `RelaxedOriginValidator`.
- A respective relaxed matcher method in the `Origin` class.
- Allows to set an origin validator on `Manager` and `DataValidator` classes. Default is still the usual `OriginValidator`.

The default `OriginValidator` is unchanged, except for being exposed to referring classes.

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/WebAuthnAuthenticationManager.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/WebAuthnAuthenticationManager.java
@@ -32,6 +32,7 @@ import com.webauthn4j.data.extension.client.AuthenticationExtensionsClientOutput
 import com.webauthn4j.util.AssertUtil;
 import com.webauthn4j.validator.AuthenticationDataValidator;
 import com.webauthn4j.validator.CustomAuthenticationValidator;
+import com.webauthn4j.validator.OriginValidator;
 import com.webauthn4j.validator.exception.ValidationException;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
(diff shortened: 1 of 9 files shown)


ynojima:
Hi, Thank you for the pull-request!
The section you cited is about "RP ID" not "Origin".
The rule for "RP ID" is relaxed for sharing a credential between origin's effective domain, and its registrable sub-domain (ex. example.com and sub.example.com), 
but the regulation for "Origin" is not relaxed. The objective of matching "Origin" is to make sure the "Origin" the user agent (browser) knows is same as what the server expected to protect users from the MITM attack.
I think there is no need to relax the regulation for "Origin". Please correct me if my understanding is wrong.
If you really need to make domain and its sub-domain valid at the same time, you may set specify them in `ServerProperty`'s constructor.

onne:
Hi Yoshikazu,
Sorry  for the late response. I understand your objection to relaxing the regulation for "Origin".
Will you be open to consider accepting a change that will allow to override the default `OriginValidator` on `*Manager` and `*DataValidator` classes? The change request will not introduce any change to existing constructors but rather an additional constructor with additional `originValidator` parameter will be introduced.
Thanks

ynojima:
Now I understand the needs for relaxing Origin validation. I'll explore how to implement it.

onne:
Following up on this. Is there anything I can do to help with this?

ynojima:
Sorry for taking time. I'll take time next week.

ynojima:
I made a pull-request #459 to resolve this issue. Could you check that meet your requirements?

ynojima:
@onne Could you check #459 whether it meets your requirements?

onne:
Close due to https://github.com/webauthn4j/webauthn4j/pull/459

PR #459, opened 2021-02-13
Expose OriginValidator to public

ynojima:
In some cases, users want to customize OriginValidator logic.
This commit make OriginValidator to interface and expose it to public.
Now users can set their own OriginValidator implementation to
RegistrationDataValidator or AuthenticationDataValidator through their
setter.


onne:
Thank you Yoshikazu. That's exactly what we needed.

onne (approved):

PR #489, opened 2021-04-07
Initialized customRegistrationValidators in contructor

inspirsea:
customRegistrationValidators was never set in the constructor.


ynojima:
Oh, Thank you for pointing out!

PR #510, opened 2021-05-15
Fix bugs using non-resident keys in WebAuthnModelAuthenticator.

JackOfMostTrades:
I was using the PackedAuthenticator today to do an integration test against a new service and ran into a few of bugs.

- COSEAlgorithmIdentifier is not an enum, so it needs a `.equals()` check instead of `==` (this comes up when it's deserialized JSON, which for me came up since I was deserializing options returned in a REST request from my service under test).
- Jackson wasn't able to serialize the PrivateKey in the `PublicKeyCredentialSource` so I switched this to a COSEKey instead.
- CipherUtil wasn't able to encrypt/decrypt the serialized credential since the attestion key was EC and not RSA, so I switched this to AES.

I added a couple of test cases that cover these things.


ynojima:
Thank you for the pull-request. I'll look into it.

PR #543, opened 2021-07-30
Add spc payment validation

davordavidovic:


ynojima:
Please resolve conflicts

webauthn4j-core/src/main/java/com/webauthn4j/validator/AuthenticationDataValidator.java
```diff
@@ -233,6 +236,47 @@ private void validate(@NonNull AuthenticationData authenticationData, @NonNull A
...
+        // validate the rpId
+        if (!Objects.equals(clientAdditionalPaymentData.getRp(), paymentAuthenticationParameters.getServerProperty().getRpId())) {
+            throw new BadRpIdException("payment rpId does not match server value");
+        }
+
+        // validate top origin
+        originValidator.validate(new Origin(clientAdditionalPaymentData.getTopOrigin()), paymentServerProperty.getOrigins());
+
+        // validate payee origin
+        originValidator.validate(new Origin(clientAdditionalPaymentData.getPayeeOrigin()), paymentServerProperty.getPayeeOrigins());
```
ynojima, line 255: Since some library users may want to validate topOrigin and payeeOrigin with custom logic rather than comparing with origin list, it is better to delegates to a new dedicated interface (`PaymentOriginValidator` ?)

webauthn4j-core/src/main/java/com/webauthn4j/data/payment/PaymentServerProperty.java
```diff
@@ -0,0 +1,40 @@
...
+import com.webauthn4j.server.ServerProperty;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Set;
+
+public class PaymentServerProperty extends ServerProperty {
+    private final PaymentCredentialInstrument instrument;
+    private final PaymentCurrencyAmount total;
+    private final Set<Origin> payeeOrigins;
```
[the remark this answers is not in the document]
davordavidovic: @ynojima what is the role of `ServerProperty` compared to `AuthenticationParameters`? In my view `payeeOrigins` is a property of the server i.e. the accepted origins on which the shopper can be during payment. Or, if we move all fields from `PaymentServerProperty` to `PaymentAuthenticationParameters` then should I also delete the `PaymentServerProperty` class itself?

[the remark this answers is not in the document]
ynojima: `ServerProperty` is a set of authentication parameters that are related to a relying party server. `payeeOrigins` is a property of a "payee", not that of a "relying party".

> should I also delete the PaymentServerProperty class itself?

Yes, it is no longer needed.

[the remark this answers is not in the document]
davordavidovic: @ynojima ah makes sense, fix submitted

webauthn4j-core/src/main/java/com/webauthn4j/data/payment/PaymentAuthenticationParameters.java
```diff
@@ -41,4 +56,8 @@ public PaymentCredentialInstrument getInstrument() {
     public PaymentCurrencyAmount getTotal() {
         return total;
     }
+
+    public Set<Origin> getPayeeOrigins() {
+        return payeeOrigins;
+    }
 }
```
ynojima, line 63: Please override `equals` and `hashCode`

webauthn4j-core/src/main/java/com/webauthn4j/data/payment/PaymentAuthenticationParameters.java
```diff
@@ -0,0 +1,80 @@
...
+        super(serverProperty, authenticator, allowCredentials, userVerificationRequired);
+        AssertUtil.notNull(payeeOrigins, "payeeOrigins must not be null");
+        AssertUtil.notNull(instrument, "payment instrument must not be null");
+        AssertUtil.notNull(total, "payment amount must not be null");
+        this.instrument = instrument;
+        this.total = total;
+        this.payeeOrigins = payeeOrigins;
+    }
+
+    public PaymentCredentialInstrument getInstrument() {
```
ynojima, line 53: Could you annotate new members like this with `@Nullable` or `NonNull`?

davordavidovic: fixed! :)

webauthn4j-core/src/main/java/com/webauthn4j/converter/CollectedClientDataConverter.java
```diff
@@ -80,6 +80,22 @@ public CollectedClientDataConverter(@NonNull ObjectConverter objectConverter) {
         }
     }
 
+    /**
+     * Converts from a byte array to {@link CollectedClientData}.
+     *
+     * @param source the source byte array to convert
+     * @return the converted object
+     */
+    public @Nullable <T extends CollectedClientData> T convert(@NonNull byte[] source, Class<T> type) {
```
ynojima, line 89:
```suggestion
    public @Nullable <T extends CollectedClientData> T convert(@NonNull byte[] source, @NonNull Class<T> type) {
```

webauthn4j-core/src/main/java/com/webauthn4j/validator/AuthenticationDataValidator.java
```diff
@@ -274,6 +313,14 @@ public void setOriginValidator(OriginValidator originValidator) {
         this.originValidator = originValidator;
     }
 
+    public PaymentOriginValidator getPaymentOriginValidator() {
```
ynojima, line 316:
```suggestion
    public @NonNull PaymentOriginValidator getPaymentOriginValidator() {
```

webauthn4j-core/src/main/java/com/webauthn4j/validator/AuthenticationDataValidator.java
```diff
@@ -274,6 +313,14 @@ public void setOriginValidator(OriginValidator originValidator) {
...
+        return paymentOriginValidator;
+    }
+
+    public void setPaymentOriginValidator(PaymentOriginValidator paymentOriginValidator) {
```
ynojima, line 320:
```suggestion
    public void setPaymentOriginValidator(@NonNull PaymentOriginValidator paymentOriginValidator) {
```

webauthn4j-core/src/test/java/com/webauthn4j/validator/PaymentOriginValidatorImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void test_successful_validation() {
+        CollectedClientAdditionalPaymentData collectedClientAdditionalPaymentData = new CollectedClientAdditionalPaymentData(
+                "example",
+                topOrigin,
+                payeeOrigin,
+                new PaymentCurrencyAmount("EUR", "15"),
+                new PaymentCredentialInstrument("Store", "favicon.ico")
+        );
+
+        target.validate(collectedClientAdditionalPaymentData, mockPaymentAuthenticationParameters());
```
ynojima, line 35:
```suggestion
        assertThatCode(()->target.validate(collectedClientAdditionalPaymentData, mockPaymentAuthenticationParameters())).doesNotThrowAnyException();
```

ynojima: This change also requires static import.

webauthn4j-core/src/test/java/com/webauthn4j/validator/PaymentOriginValidatorImplTest.java
```diff
@@ -0,0 +1,85 @@
...
+    void test_successful_validation() {
+        CollectedClientAdditionalPaymentData collectedClientAdditionalPaymentData = new CollectedClientAdditionalPaymentData(
+                "example",
+                topOrigin,
+                payeeOrigin,
+                new PaymentCurrencyAmount("EUR", "15"),
+                new PaymentCredentialInstrument("Store", "favicon.ico")
+        );
+
+        assertThatCode(()->target.validate(collectedClientAdditionalPaymentData, mockPaymentAuthenticationParameters())).doesNotThrowAnyException();
```
ynojima, line 35: This change also requires static import.

webauthn4j-core/src/main/java/com/webauthn4j/data/payment/PaymentCurrencyAmount.java
```diff
@@ -0,0 +1,51 @@
...
+        if (o == null || getClass() != o.getClass()) return false;
+        PaymentCurrencyAmount that = (PaymentCurrencyAmount) o;
+
+        if (!currency.equals(that.currency)) return false;
+
+        try {
+            return Double.parseDouble(value) == Double.parseDouble(that.value);
+        } catch (NumberFormatException e) {
+            return false;
+        }
```
ynojima, line 44: This makes following code false:
```
new PaymentCurrencyAmount("USD","invalid").equals(new PaymentCurrencyAmount("USD","invalid"))
```
I think it is inappropriate. Verification should be conducted at the designated place (`AuthenticationDataValidator#validatePaymentAmount`).

davordavidovic: amount should be a valid number, so your example should return false imo. Alternatively we can throw a numberformat exception to handle this case?

ynojima: I think `new PaymentCurrencyAmount("USD","invalid").equals(new PaymentCurrencyAmount("USD","invalid"))` should return `true`. 
Even though the value is invalid from the point of `PaymentCurrencyAmount` design, `equals` method need to just focus on value equality.

davordavidovic: ok i see your point, i will then extract the extra logic from the `equals` to the `AuthenticationDataValidator#validatePaymentAmount` method

ynojima:
Thank you for addressing the issues! merged into the `spc-dev`.

PR #552, opened 2021-08-07
Add hmac-secret extension data classes

ynojima:


ynojima:
Duplication is by design. Merging.

Issue #564, opened 2021-08-28
MDS3 support


ynojima:
**Is your feature request related to a problem? Please describe.**

Support FIDO MDS3
https://fidoalliance.org/metadata/

**Describe the solution you'd like**

Implement MDS3 support library and deprecate the MDS2 support. 

**Additional context**

https://groups.google.com/a/fidoalliance.org/g/fido-dev/c/KTXB4NWAp5U/m/cECm7DGpAAAJ?utm_medium=email&utm_source=footer&pli=1

choojunyup:
I wish the webauthn4j project could support both md2 and md3.
Thank you for your interest in this issue.

ynojima:
MDS3 Support is released as part of [0.19.0.RELEASE ](https://github.com/webauthn4j/webauthn4j/releases/tag/0.19.0.RELEASE)

PR #597, opened 2021-11-07
Add fido registry predefined values

ynojima:
Add fido registry predefined values (https://fidoalliance.org/specs/common-specs/fido-registry-v2.1-ps-20191217.html) to the webauthn4j-core module.
Some of them are already defined in the webauthn4j-metadata module, and the are moved to the webauthn4j-core module too.

Breaking change:
- Some types defined in webauthn4j-metadata module as fido registry predefined values are moved to the webauthn4j-core module.


ynojima:
This is a preparation for #564

PR #611, opened 2021-12-02
Correct the description error in the Japanese translation

kuraun:
Nice to meet you. Thank you for your wonderful work.
By the way, when I read the document (I'm Japanese), there was something wrong with the description, and I thought it was incorrect, so I presented a correction plan.
We apologize for the inconvenience, but thank you for your understanding.

Best regards,


ynojima:
Good catch! Thank you for correction. Will merge when it passes the CI.

PR #801, opened 2023-07-05
Adding nullable to Options inside AuthenticatorGetInfo

billlyd123:
This will resolve an issue with Serialization/ Deserialization when only some of the options are present. Discussion post here: https://github.com/webauthn4j/webauthn4j/discussions/797


ynojima:
Sorry for overlooking the topic in the Discussions.  I'm on vacation and will check no later than the 18th.

ynojima:
Thank you for your fix. Now it is merged into master.

billlyd123:
Do you know when the next release will be?

ynojima:
WebAuthn4J is released on the 1st of every month. So, next release will be  Aug. 1st.

PR #863, opened 2023-12-30
Add CompoundAttestationStatement

ynojima:


ynojima:
As CompoundAttestationStatement is not supported by Chrome, this pull-request is pending.

Issue #845, opened 2023-11-30
Remove Serializable from data classes as it is misleading


ynojima:
WebAuthn4j data classes implements `Serializable`, but they don't assure serialized data compatibility between versions.
As implementing `Serializable` interface is misleading, remove `Serializable` interface.

Ref: https://github.com/webauthn4j/webauthn4j/issues/844

PR #885, opened 2024-03-09
Remove Serializable

ynojima:
Since webauthn4j data classes are not designed to be serialized with ObjectOutputStream, Serializable is removed.
Please use JsonConverter or CborConverter instead.

Ref: https://github.com/webauthn4j/webauthn4j/issues/845


ynojima:
This duplication is expected.

PR #914, opened 2024-06-05
Fix URL for Apache license

stianst:
`http://www.apache.org/license/LICENSE-2.0.txt` returns 404, the correct link is `https://www.apache.org/licenses/LICENSE-2.0.txt`


ynojima:
Oh, Thank you for the fix.

Issue #909, opened 2024-05-30
Add Non-Blocking API 


ynojima:
The current webauthn4j API is a blocking API. We will add a non-blocking API that can be utilized from Quarkus' `HttpAuthenticationMechanism` or `IdentityProvider` implementations.

maxandersen:
from a chat/call 2024-06-05 :

- introduce http client SPI to allow hook in another http client vert.x
- add non-blocking api in collaboration with updates coming from vertx module update

ynojima:
Design doc: https://github.com/webauthn4j/webauthn4j/wiki/WebAuthn4J-Reactive-Design-Doc
WIP pull-request: https://github.com/webauthn4j/webauthn4j/pull/920

rwinch:
I'm also interested in this for Spring Security's Reactive support.

ynojima:
`webauthn4j-core-async` is now available in maven central.

https://github.com/webauthn4j/webauthn4j/releases/tag/0.27.0.RELEASE
https://central.sonatype.com/artifact/com.webauthn4j/webauthn4j-core-async

FroMage:
Thanks a lot!

PR #920, opened 2024-06-09
New feature: Async support

ynojima:
Design doc: https://github.com/webauthn4j/webauthn4j/wiki/WebAuthn4J-Async-Design-Doc
Issue: https://github.com/webauthn4j/webauthn4j/issues/909

Snapshot build for this pull-request is available at jitpack.io.
You may try with this:

Repository:
```
	<repositories>
		<repository>
		    <id>jitpack.io</id>
		    <url>https://jitpack.io</url>
		</repository>
	</repositories>
```

Dependency:
```
	<dependency>
	    <groupId>com.github.webauthn4j.webauthn4j</groupId>
	    <artifactId>webauthn4j-core-async</artifactId>
	    <version>reactive-support-SNAPSHOT</version>
	</dependency>
```


FroMage (approved):
That looks fine :) Note that I did not have time yet to try this out, unfortunately :(

ynojima:
This is still WIP status, not ready to be tested, but I just want to show I'm working on this :)

ynojima:
Passed all mandatory FIDO conformance test cases
image

FroMage:
OK great. I'm working on porting vertx-webauthn4j to this ATM. I'll let you know when I have questions or news :)

FroMage:
Hi,

I'm making some progress porting `vertx-auth-webauthn` to this. I'm not happy with the existing tests in Vert.x, they're fairly light and few.

But one of the test is using an expired certificate, to make sure we reject it. The defaults of `WebAuthnRegistrationAsyncManager.createNonStrictWebAuthnRegistrationAsyncManager` do not check any certificate, so I've replaced it with a call to check their validity, hoping I'd get to notice it's expired, but I'm not getting a validity check but a trust chain failure:

```
Caused by: com.webauthn4j.verifier.exception.CertificateException: invalid cert path
	at webauthn4j.core.async.reactive.support@0.9.9.RELEASE-gda14d08-1525/com.webauthn4j.async.verifier.attestation.trustworthiness.certpath.DefaultCertPathTrustworthinessAsyncVerifier.verifyCertPath(DefaultCertPathTrustworthinessAsyncVerifier.java:107)
	at webauthn4j.core.async.reactive.support@0.9.9.RELEASE-gda14d08-1525/com.webauthn4j.async.verifier.attestation.trustworthiness.certpath.DefaultCertPathTrustworthinessAsyncVerifier.lambda$verify$1(DefaultCertPathTrustworthinessAsyncVerifier.java:68)
	at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
	... 44 more
Caused by: java.security.cert.CertPathValidatorException: Path does not chain with any of the trust anchors
	at java.base/sun.security.provider.certpath.PKIXCertPathValidator.validate(PKIXCertPathValidator.java:157)
	at java.base/sun.security.provider.certpath.PKIXCertPathValidator.engineValidate(PKIXCertPathValidator.java:83)
	at java.base/java.security.cert.CertPathValidator.validate(CertPathValidator.java:309)
	at webauthn4j.core.async.reactive.support@0.9.9.RELEASE-gda14d08-1525/com.webauthn4j.async.verifier.attestation.trustworthiness.certpath.DefaultCertPathTrustworthinessAsyncVerifier.verifyCertPath(DefaultCertPathTrustworthinessAsyncVerifier.java:103)
```

This is trying to validate a certificate with a single packed attestation certificate for `CN=Yubico U2F EE Serial 1213939126, OU=Authenticator Attestation, O=Yubico AB, C=SE` which is not a root certificate.

I suppose this test is probably invalid because it should fail anyway since that certificate does not have a complete chain, but I wanted to check with you that certificate validity dates are checked in webauthn4j. I'm sure it's the case, but worth checking :)

Also, I'm not sure how to plug in the list of CRL for validation, could you advise?

ATM I'm building my manager like this:

```java
    TrustAnchorAsyncRepository something = new KeyStoreTrustAnchorAsyncRepository(keyStore);
    webAuthnManager = new WebAuthnAsyncManager(
            Arrays.asList(
                    new NoneAttestationStatementAsyncVerifier(),
                    new FIDOU2FAttestationStatementAsyncVerifier(),
                    new PackedAttestationStatementAsyncVerifier(),
                    new TPMAttestationStatementAsyncVerifier(),
                    new AndroidKeyAttestationStatementAsyncVerifier(),
                    new AndroidSafetyNetAttestationStatementAsyncVerifier(),
                    new AppleAnonymousAttestationStatementAsyncVerifier()
            ),
            new DefaultCertPathTrustworthinessAsyncVerifier(something),
            new DefaultSelfAttestationTrustworthinessAsyncVerifier(),
            new ObjectConverter()

```

FroMage:
Also, does webauthn4j support https://fidoalliance.org/metadata/ ? The docs say yes, but does not give an example of how to use it.

And last question (for today): what are the async parts in validation? Are you making remote requests? If yes, using what transport?

ynojima:
Certificate dates validity is checked by `DefaultCertPathTrustworthinessAsyncVerifier`. It calls JDK's `CertPathValidator` to checks the certificate chain and the certificate's validity dates. 
https://github.com/webauthn4j/webauthn4j/blob/955ba688c11b5f97cd707cc8c44b8965813093d9/webauthn4j-core-async/src/main/java/com/webauthn4j/async/verifier/attestation/trustworthiness/certpath/DefaultCertPathTrustworthinessAsyncVerifier.java#L103

Please note `CertPathValidator`'s revocation check is explicitly disabled because JDK's `RevocationChecker` fetches CRL from a CRL distribution point with blocking API.
https://github.com/webauthn4j/webauthn4j/blob/955ba688c11b5f97cd707cc8c44b8965813093d9/webauthn4j-core-async/src/main/java/com/webauthn4j/async/verifier/attestation/trustworthiness/certpath/DefaultCertPathTrustworthinessAsyncVerifier.java#L83-L85

I assume vertx-auth-webauthn takes CRL as an argument to avoid downloading CRL, but `webauthn4j-core-async` disabled attestation certificate's CRL check at this moment since CRL check is not mandatory in WebAuthn spec because FIDO Metadata service provides revocation informantion (FIDO conformance test can be passed without CRL check).

If required for compatibility with vertx-auth-webauthn, I can modify the `DefaultCertPathTrustworthinessAsyncVerifier` to accept a CRL and perform the revocation check using it.

> Also, does webauthn4j support https://fidoalliance.org/metadata/ ?

Yes, webauthn4j supports it. Please use `MetadataBLOBBasedTrustAnchorAsyncRepository` in conjunction with `FidoMDS3MetadataBLOBAsyncProvider`.
`MetadataBLOBBasedTrustAnchorAsyncRepository` can be used as a `TrustAnchorAsyncRepository`, and it exists in `webauthn4j-metadata-async`.
I'll share you a sample code later.

In FIDO Metadata service processing, unlike attestation certificate processing, revocation check is mandatory when verifying the downloaded metadata BLOB. 
I use a non-blocking HTTP client here to download the CRL from the CRL distribution point specified in the certificate, and perform validation using the CRL.
https://github.com/webauthn4j/webauthn4j/blob/da14d080227d1b5588e77fb8989182390f4cd378/webauthn4j-metadata-async/src/main/java/com/webauthn4j/async/metadata/FidoMDS3MetadataBLOBAsyncProvider.java#L132-L248
This is the asynchronous part.

JDK's standard `HttpClient` is used as a non-blocking HTTP client, but you can replace it by injecting through `HttpAsyncClient` interface.
https://github.com/webauthn4j/webauthn4j/blob/955ba688c11b5f97cd707cc8c44b8965813093d9/webauthn4j-metadata-async/src/main/java/com/webauthn4j/async/metadata/SimpleHttpAsyncClient.java#L16-L27

FroMage:
> Certificate dates validity is checked by DefaultCertPathTrustworthinessAsyncVerifier. It calls JDK's CertPathValidator to checks the certificate chain and the certificate's validity dates.

Yes, indeed. I just verified that.

The current Vert.x/webauthn implementation checked the validity period, and that every certificate in the chain had a valid signature. But it did not check that the root CA was known. So the implementations differ here, because as far as I can see, in webauthn4j, if I want to check the validity and signatures, I also get a validation error that the root CA is unknown.

Also, I had to manually load the JDK's keystore for Root CAs, I did not find any way to get the `KeyStore` instance already loaded by the JDK, that's a bit annoying because I know it must have it loaded somewhere.

And there are some CAs that are added manually in Vert.x/webauthn: https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/WebAuthnOptions.java#L50

Three for AndroidKey/AndroidSafetynet/Apple attestations.

One for the metadata service.

Do I need to set them up anywhere or are they already known to webauthn4j?

> I assume vertx-auth-webauthn takes CRL as an argument to avoid downloading CRL

Yes, they currently come from configuration.

ynojima:
First, allow me to explain my understanding of the authenticator's attestation.

The attestation is data that proves whether the authenticator is a trusted authenticator model. For a trusted authenticator, the manufacturer issues and embeds an attestation certificate and its private key.
The trustworthiness of the authenticator can be determined by checking that the attestation certificate is chained to the root attestation certificate published by the  manufacturer of the authenticator.
https://fidoalliance.org/fido-technotes-the-truth-about-attestation/

However, since the attestation data can be used to track the model of the authenticator, there are privacy concerns. Outside of enterprise use cases where strong security is required, it is generally not recommended to request or verify attestation.
Therefore, when creating a credential with the WebAuthn API, the default option is to not to request attestation.
https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#enumdef-attestationconveyancepreference

A scenario where attestation is still required would be if you need to strictly verify that users are only using trusted authenticators, and you can enforce the use of such trusted authenticators on users.
This is not for consumer-facing services but rather for enterprise services, such as employee identity provider for internal systems that require high security.
In enterprise scenarios, the IT department could distribute certified security keys to employees. In such cases, the attestation verification policy on the authentication server would only allow the models of these security keys.
If strict enforcement isn't necessary, allowing only FIDO Alliance-certified authenticators could also be an option.

This means that services implementing WebAuthn login can have one of these policies:

1. Allow any type of authenticator
2. Allow FIDO Alliance certified authenticators only
3. Allow organization approved authenticators only

The attestation verification must be conducted based on the selected policy.

In WebAuthn4J, for consumer-facing services where any authenticator is allowed, I provide `WebAuthnAsyncManager.createNonStrictWebAuthnAsyncManager()`, which does no attestation verification.

To verify the attestation, the root attestation certificates of trusted authenticators can be provided through the `TrustAnchorAsyncRepository`, which is passed during the instantiation of `WebAuthnAsyncManager`.

If you want to allow only FIDO Alliance certified authenticators, you can configure the combination of `MetadataBLOBBasedTrustAnchorAsyncRepository` and `FidoMDS3MetadataBLOBAsyncProvider`.
If you want to allow only organization approved authenticators, you can, for example, use `KeyStoreTrustAnchorAsyncRepository` to provide their root certificates from a keystore.

Now, based on the above, I'll respond to your comment.

>Do I need to set them up anywhere or are they already known to webauthn4j?

I do not include well-known root certificates in webauthn4j.
For attestation verification, we anticipate use cases where it is important to strictly choose which models to trust. 
Therefore, webauthn4j expects users to set up trusted root certificates by themselves. Another reason to avoid inclusion is the difficulty of keeping up with updates of root certificates.

WebAuthn4J does not trust certificates stored in the JDK default keystore. The current Vert.x/webauthn implementation also does not seem to trust certificates stored in the default JDK keystore as root attestation certificates.

>The current Vert.x/webauthn implementation checked the validity period, and that every certificate in the chain had a valid signature. But it did not check that the root CA was known. 

I have checked the code of the current Vert.x/webauthn implementation, and it does check whether the root CA is known. However, it always uses the list of root CA certificates returned by FIDO MDS as the trust anchors.
https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/attestation/PackedAttestation.java#L130-L133

If you have performed verification of the attestation `CN=Yubico U2F EE Serial 1213939126, OU=Authenticator Attestation, O=Yubico AB, C=SE,`, it passed because the root attestation certificate of Yubikey is listed in the FIDO MDS.

If you want to align webauthn4j with the Vert.x/webauthn implementation, you should use the combination of `MetadataBLOBBasedTrustAnchorAsyncRepository` and `FidoMDS3MetadataBLOBAsyncProvider` as the `TrustAnchorAsyncRepository`.

However, for use cases where only organisation-approved authenticators are allowed, I think the current Vert.x/webauthn implementation, which always allows authenticators listed in the FIDO MDS, does not provide enough flexibility for library users.

FroMage:
Thanks a lot for your explanation. That's very clear and helpful.

> I have checked the code of the current Vert.x/webauthn implementation, and it does check whether the root CA is known. However, it always uses the list of root CA certificates returned by FIDO MDS as the trust anchors.
https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/attestation/PackedAttestation.java#L130-L133

That is not my understanding of the code. What I think it does, judging by who calls `WebAuthnOptions.getRootCertificate()` I see it called for apple and android attestations, either adding it at the end of the chain (apple) or making sure that it's the last in the chain of attestations.

The other place where they're used is for the `mds` one and only to verify that the FIDO MDS was valid, but that metadata call is optional. If that metadata call is fetched and added to the system, then only the attestations with a matching `aaguid` will be checked for their root certificates as provided by the metadata: https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/metadata/MetaData.java#L190

For https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/attestation/PackedAttestation.java#L130-L133 in my opinion the only things that are checked (if not metadata is loaded from MDS) are certificate validity, that they are chained properly, signatures are correct, and the last one is self-signed. I could not find anywhere that it checks that the last one has to be a root CA. Perhaps I missed it.

But I do agree that most users will not want to check attestations. In fact, I see Vert.x has an option for https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#enumdef-attestationconveyancepreference at https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/WebAuthnOptions.java#L224 defaulting to `NONE` but it does not appear to be used to determine whether to check the attestations or not. It seems only used in https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-webauthn/src/main/java/io/vertx/ext/auth/webauthn/impl/WebAuthnImpl.java#L219 to create credential options.

I wonder if we need an additional configuration option to determine whether we need to check attestations, or if I can use that, and only check attestations for the `ENTERPRISE` value?

In any case, the current Vert.x implementation regarding attestations seems to be:

- always check them if provided
- checking their validity
- checking that they chain properly, that signatures are valid
- only check that the last one is apple/android for those attestation formats
- check that the last one corresponds to one provided by MDS if used, and if a matching AAGUID is found
- otherwise never check the last one

This seems to be different from the options I have with webauthn4j. A bit more (validity date, and apple/android root CAs) in the case of `createNonStrictWebAuthnAsyncManager` and less than the strict version (root CAs are never checked outside of apple/android or MDS with matching AAGUID).

I'm not saying we should change anything in webauthn4j. I'm just hilighting the difference I found so you can tell me what to do :)

ynojima:
Thank you for the explanation. 
I now understand that there are cases where the current vert.x/webauthn implementation doesn't check whether the root certificate is known. 
I overlooked that even if `metadata` is passed to `Attestation#validate`, unless `Metadata#loadMetadata` is not called, `metadata` doesn't contain any `MetaDataEntry` fetched from MDS.
I also overlooked that the current vert.x/webauthn implementation doesn’t throw exception when there is no MetaDataEntry matching the AAGUID.

To be honest, I now think the current behavior and API signatures of the Vert.x/WebAuthn implementation are not good. 
Since there is a case where the root certificate is not checked, the current vert.x/webauthn implementation doesn't fulfill the purpose of attestation checks, which limit authenticators to trusted ones.
While "check exempted" mode is necessary for non-enterprise use-cases, it should also have a mode that enforces the check.

It is also an issue that `WebAuthnOptions` accepts only one root certificate per format, as it prevents supporting cases where multiple authenticator models need to be trusted.

FroMage:
I can't seem to be able to download the `webauthn4j-metadata-async` module, this doesn't work:

```xml
	<dependency>
	    <groupId>com.github.webauthn4j.webauthn4j</groupId>
	    <artifactId>webauthn4j-metadata-async</artifactId>
	    <version>reactive-support-SNAPSHOT</version>
	</dependency>
```

Do you know why?

ynojima:
I'm not sure why, but it seems latest reactive-support branch build failed in jitpack.io: https://jitpack.io/com/github/webauthn4j/webauthn4j/reactive-support-0.9.9.RELEASE-gfe393e7-1526/build.log

Since the previous build succeeded(https://jitpack.io/#webauthn4j/webauthn4j/reactive-support-0.9.9.RELEASE-gda14d08-1525),
Could you try with this for workaround?
```
	<dependency>
	    <groupId>com.github.webauthn4j.webauthn4j</groupId>
	    <artifactId>webauthn4j-core</artifactId>
	    <version>reactive-support-0.9.9.RELEASE-gda14d08-1525</version>
	</dependency>
```

FroMage:
Doesn't work due to transitive dependency being on another version:

```
[ERROR] Failed to execute goal on project vertx-auth-webauthn: Could not resolve dependencies for project io.vertx:vertx-auth-webauthn:jar:5.0.0-SNAPSHOT
[ERROR] dependency: com.github.webauthn4j.webauthn4j:webauthn4j-metadata:jar:reactive-support-SNAPSHOT (compile)
[ERROR] 	Could not find artifact com.github.webauthn4j.webauthn4j:webauthn4j-metadata:jar:reactive-support-0.9.9.RELEASE-g3efc9d0-1500 in jitpack.io (https://jitpack.io)
```

ynojima:
I rerun the jitpack build by pushing a commit. Please try `mvn` command again with `-U` option to force fetching the `reactive-support-SNAPSHOT`.

FroMage:
Thanks, I just compiled it locally it's easier :)

FroMage:
BTW, I think you can merge this, it seems to be working just fine :)

FroMage (approved):

ynojima:
> Do you think I should validate the attestations in the case of direct, indirect and enterprise or should I make validation an option unrelated to the AttestationConveyancePreference?

Yes, I think it's a good idea to link `AttestationConveyancePreference` value with whether to validate or not.

> 1. when we look up credentials in the DB we sometimes look for `credentialId` and sometimes for `userName`

Yes, credentialId and username must be supported as a look up key.
WebAuthn relying party need to look up a single credential record from a DB with a `credentialId` when verifying a credential record.
WebAuthn relying party also needs to look up list of credential records from a DB with `username` when building a `PublicKeyCredentialCreationOptions` or a `PublicKeyCredentialRequestOptions`.
`excludeCredentials` of `PublicKeyCredentialCreationOptions` takes a list of credentialIds. This ensures that the new credential is not created on an authenticator that already contains a credential mapped to this user account.
WebAuthn has a type of credential named ["Server-side Credential"](https://w3c.github.io/webauthn/#server-side-credential), which requires WebAuthn JS API caller to supply credential id(s) to `allowCredentials` list of `PublicKeyCredentialRequestOptions`.
(Credential id need to be supplied because a kind of seed of credential private key is encoded in the `credentialId`. It's concept is iherited from FIDO U2F. For more details, please refer [a guide from Yubico](https://web.archive.org/web/20190712075231/https://developers.yubico.com/U2F/Protocol_details/Key_generation.html).)
(Client side discoverable credentials doesn't need `allowCredentials`)

> in both cases it turns a list, because a single user can have multiple credentials stored, and multiple users can share the same authenticator, 

a single user account can have multiple authenticators, but since [`credentialId` is an unique identifier](https://w3c.github.io/webauthn/#credential-id) for one credential record, single credential record must be returned when we look up with `credentialId`.

> in the case of a yubikey shared on a home computer between family members for example (though this last example sounds dodgy in my opinion, because then every family member can log in each other's account simply by knowing the other member's user names)

Even if an authenticator is shared between multiple user (account)s, it doesn't mean a `credentialId` is associated with multiple credential records.
I couldn't understand why quarkus-security-webauthn's `WebAuthnUserProvider#findWebAuthnCredentialsByCredID` returns list of authenticator.

ynojima:
Thank you for the approval. As there's no immediate need to rush the merge, I would like to wait for a while longer (some issume may be found during the vert.x integration). 
If no issues are found, I would like to merge :)

FroMage:
No issues have been found so far, and I can't make a Vert.x or Quarkus PR without a release of this, so go ahead :)

Minor nitpick, I've been hit by exceptions such as this one:

```
            throw new NotAllowedAlgorithmException("alg not listed in options.pubKeyCredParams is used.");
```

From a user-experience perspective, it's always much nicer to include actionable information in the exceptions. Reading this, the user can probably figure out what the list of `options.pubKeyCredParams` is, because it's configuration, but there's no way users can figure out what is the forbidden algo without using a debugger, because it's so hidden in the response via base64/json/cbor.

It would be more useful to mention in the exception message the algo that is being forbidden, so users can decide whether to allow it or not.

ynojima:
> Well, I suppose it depends on how the credentialId is generated.

Authenticator is need to generate unique `credentialId`. Furthermore, the Relying Party should reject registration when 
 an existing `credentialId` is provided. So the database should have unique constraint on the `credentialId` field.

Please check step 26 of "7.1. Registering a New Credential" in the webauthn specification.
> Verify that the [credentialId](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#authdata-attestedcredentialdata-credentialid) is not yet registered for any user. If the [credentialId](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#authdata-attestedcredentialdata-credentialid) is already known then the [Relying Party](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#relying-party) SHOULD fail this [registration ceremony](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#registration-ceremony).
>
> NOTE: The rationale for [Relying Parties](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#relying-party) rejecting duplicate [credential IDs](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#credential-id) is as follows: [credential IDs](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#credential-id) contain sufficient entropy that accidental duplication is very unlikely. However, [attestation types](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#attestation-type) other than [self attestation](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#self-attestation) do not include a self-signature to explicitly prove possession of the [credential private key](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#credential-private-key) at [registration](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#registration) time. Thus an attacker who has managed to obtain a user’s [credential ID](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#credential-id) and [credential public key](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#credential-public-key) for a site (this could be potentially accomplished in various ways), could attempt to register a victim’s credential as their own at that site. If the [Relying Party](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#relying-party) accepts this new registration and replaces the victim’s existing credential registration, and the [credentials are discoverable](https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#discoverable-credential), then the victim could be forced to sign into the attacker’s account at their next attempt. Data saved to the site by the victim in that state would then be available to the attacker.

https://www.w3.org/TR/2023/WD-webauthn-3-20230927/#reg-ceremony-assess-trust

ynojima:
### Exception message

Yes, exception message needs improvement... Thank you for suggestion.

### Release

I’ve added a note declaring that `webauthn4j-core-async` and `webauthn4j-metadata-async` are in an experimental status. 
I will now merge this. The note will be removed when the Vert.x or Quarkus side is released.
Next webauthn4j release will be Oct. 1st.

FroMage:
Perfect, thanks a lot!

ynojima:
`webauthn4j-core-async` is now available in maven central.

https://github.com/webauthn4j/webauthn4j/releases/tag/0.27.0.RELEASE
https://central.sonatype.com/artifact/com.webauthn4j/webauthn4j-core-async

PR #964, opened 2024-08-28
Add RegistrationResponseJSON and AuthenticationResponseJSON support

ynojima:


ynojima:
Follow up change: https://github.com/webauthn4j/webauthn4j/pull/969

Issue #989, opened 2024-09-24
Declarative JPMS support


vietj:
**Is your feature request related to a problem? Please describe.**

Webauthn4j does not support Java modules. Although it might be possible to add support for named automatic modules, it would be better to support explicit modules (with a module-info.class descriptor). 

**Describe the solution you'd like**

Add module-info declarations describing the modularity of the library.

**Describe alternatives you've considered**

None.

**Additional context**

- Vert.x 5 might use webauthn4j and provides support for JPMS, having explicit modules makes definitely a difference.
- There might be some issues for this task: the Apache Kerby ASN1 parser does not provide module support and  a workable solution should be found
- I can contribute support for this, if needed or help testing the JPMS aspect . I did bring JPMS support to the Netty library (https://github.com/netty/netty/issues/14176)

ynojima:
Automatic module name will be added in [the next release](https://github.com/webauthn4j/webauthn4j/pull/988),
but yes, Apache Kerby ASN1 parser is the issue for making webauthn4j named module.

This could be a good chance to implement its own ASN.1 DER parser using [vertx-auth asn1 parser](https://github.com/eclipse-vertx/vertx-auth/blob/master/vertx-auth-common/src/main/java/io/vertx/ext/auth/impl/asn/ASN1.java) as a reference and remove the dependency on Apache Kerby. I'll look into it.

vietj:
thanks for the responsiveness @ynojima copying the vertx parser is what I meant with "workable solution"

ynojima:
I'm working on replacing ASN.1 parser (https://github.com/webauthn4j/webauthn4j/pull/1004). Please wait a little longer.

ynojima:
@vietj I've made a pull-request to support JPMS: https://github.com/webauthn4j/webauthn4j/pull/1005
I confirmed it works with [a modularized app](https://github.com/webauthn4j/quarkus-security-passkeys-poc/pull/1/files), but this is my first Java module, could you review it meets your requirement?

vietj:
@ynojima I'll have a look

vietj:
I think we can integrate this in vertx-auth-webauthn4j and write test using the module path to check it works correctly

vietj:
good job @ynojima thank you!

ynojima:
Thank you for response!

The artifact is published to `https://oss.sonatype.org/content/repositories/snapshots` repository.
Please use following cordinate when performing the test.
```
        <dependency>
[...]

PR #1005, opened 2024-10-10
Add explicit java module support

ynojima:
Addresses https://github.com/webauthn4j/webauthn4j/issues/989

- Exports all packages except internal ones

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,6 +147,7 @@ subprojects {
         options.compilerArgs << "-Xlint:static"
         //options.compilerArgs << "-Xlint:deprecation" //Temporary suspended to allow com.webauthn4j.metadata.legacy classes TODO: enable this option again when legacy classes are removed
         options.compilerArgs << "-Werror"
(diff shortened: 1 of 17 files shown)


ynojima:
It is confirmed a modularized app now can use webauthn4j as an explicit module: [poc application](https://github.com/webauthn4j/quarkus-security-passkeys-poc/pull/1/files)

ynojima:
Thank you for your review!

> If you do proper explicit java modules, you do not have to do the "automatic modules" hack in the jar manifest that I asked you about earlier, please undo the earlier patch.

"automatic modules" configuration will be removed with this pull-request: https://github.com/webauthn4j/webauthn4j/pull/1005/files#diff-9513e757bcce6c631200c1e17f33c8a8f06ec8b41bcb6a32f6e2e7b9b7dd0b7f

> With regard to "requires" vs "requires transitive" you need to decide if you want to pull in jackson jars recursively, or expect the user of webathn4j to explicitly require jackson modules when using webauthn4j

Added a new commit 0d43fb790aa3c817767f5d5d23c3812d0bb9288a, Thank you!

PR #1007, opened 2024-10-10
align ServerProperty's toString() to other toString() implementations

grossjonas:
... by adding a comma before the next field


ynojima:
Good catch! Thank you for your pull-request.

PR #1044, opened 2024-11-12
Rewrite documentation

ynojima:
updated the document to introduce the WebAuthn authentication flow first,
then introduce how to implement with webauthn4j


ynojima:
Follow-up issue: https://github.com/webauthn4j/webauthn4j/issues/1045

PR #1081, opened 2025-01-01
Redesign HttpClient to align with HttpAsyncClient

ynojima:
Redesign `HttpClient#fetch` to return `Response` instead of `String`. 

This only affects if you have implemented your own `HttpClient`. `HttpAsyncClient` is not affected.


ynojima:
Code coverage is expected. Merging.

PR #1092, opened 2025-02-15
Bugfix: Deserialization issue with HMACGetSecretInput and HMACGetSecretOutput

### LINKED ISSUE #1091: Deserialization issue with HMACGetSecretInput and HMACGetSecretOutput
Hi,

I have two question around extension in the option response.

1.) When having extensions in the `PublicKeyCredentialRequestOptions` or `PublicKeyCredentialCreationOptions`
The type `AuthenticationExtensionsClientInputs` is used to house the extensions.

Since all of these types are having Json related annotations from Jackson my assumption these are intended to be used for serialization.

However the type `HMACGetSecretInput` used here: https://github.com/webauthn4j/webauthn4j/blob/master/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/client/AuthenticationExtensionsClientInputs.java#L43

Does not have a public constructor or any other annotation which would be required to make serializable by Jackson.
I also don't seem to find any specific deserializes written for it.

So creating something like this:
```
AuthenticationExtensionsClientInputs<AuthenticationExtensionClientInput> extensions = new AuthenticationExtensionsClientInputs.BuilderForAuthentication()
                .setAppid("testappid")
                .setAppidExclude("testappidexclude")
                .setHMACGetSecret(new HMACGetSecretInput(new byte[] {0}, new byte[] {1}))
                .setUvm(Boolean.TRUE)
                .set("myextension", "test")
                .build();

        ServerPublicKeyCredentialGetOptionsRequest serverPublicKeyCredentialGetOptionsRequest = ServerPublicKeyCredentialGetOptionsRequest.builder()
                .username("")
                .userVerification(USER_VERIFICATION_REQUIREMENT)
                .extensions(extensions)
                .build();
        String json = objectMapper.writeValueAsString(serverPublicKeyCredentialGetOptionsRequest);
```
Produces the following which is likely already wrong:
```
{"username":"","userVerification":"required","extensions":{"appid":"testappid","appidExclude":"testappidexclude","uvm":true,"credProps":null,"credentialProtectionPolicy":null,"enforceCredentialProtectionPolicy":null,"hmacCreateSecret":null,"hmacGetSecret":{"salt1":"AA","salt2":"AQ"},"myextension":"test"}}
```

Than parsing back this to an object by Jackson triggers the following exception:
```
Caused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `com.webauthn4j.data.extension.HMACGetSecretInput` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

Am I overlooking something or this is indeed a bug?

2.) I discovered several inconsistencies among the property names in `AuthenticationExtensionsClientInputs` and the IANA extension database: https://www.iana.org/assignments/webauthn/webauthn.xhtml
- `credentialProtectionPolicy` vs `credProtect` (?)
- `enforceCredentialProtectionPolicy` is this related to the `credProtect` or something else?
- `hmacCreateSecret` vs `hmac-secret`
- `hmacGetSecret` vs `hmac-secret`

Are those incorrect or simply some custom extension by the project and they have nothing to do with the ones on the IANA list?
(I didn't dive into the specification for each of those separately)

Version **0.28.5.RELEASE**
Using Jackson **2.18.2**

Thanks a lot,
Adam

ynojima:
Closes #1091

### DIFF
--- a/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/HMACGetSecretInput.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/HMACGetSecretInput.java
@@ -16,6 +16,8 @@
 
 package com.webauthn4j.data.extension;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.webauthn4j.util.ArrayUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -24,10 +26,15 @@ import java.util.Arrays;
 
 public class HMACGetSecretInput {
 
+    @JsonProperty
     private final byte[] salt1;
+    @JsonProperty
     private final byte[] salt2;
 
-    public HMACGetSecretInput(@NotNull byte[] salt1, @Nullable byte[] salt2) {
+    @JsonCreator
+    public HMACGetSecretInput(
+            @NotNull @JsonProperty("salt1") byte[] salt1,
+            @Nullable @JsonProperty("salt2") byte[] salt2) {
         this.salt1 = ArrayUtil.clone(salt1);
         this.salt2 = ArrayUtil.clone(salt2);
     }
--- a/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/HMACGetSecretOutput.java
+++ b/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/HMACGetSecretOutput.java
@@ -16,6 +16,8 @@
 
 package com.webauthn4j.data.extension;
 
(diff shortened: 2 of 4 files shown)


adam-siklosi:
Thanks for the detailed explanation & the fix!

An other quick question: BuilderForAuthentication in the AuthenticationExtensionsClientInputs contains most of the builder methods with the @NotNull annotation. Though my understanding is that all extension are optional.
https://github.com/webauthn4j/webauthn4j/blob/master/webauthn4j-core/src/main/java/com/webauthn4j/data/extension/client/AuthenticationExtensionsClientInputs.java#L295-L305

Am I overlooking something?

PR #1101, opened 2025-03-05
Update README.md with OpenUnison as a service that uses webauthn4j

mlbiam:
OpenUnison has been using WebAuthn4j for a couple of years now.  our largest implementation has over 10k passkeys registered!


ynojima:
Thank you for letting me know that you are using WebAuthn4J. I have placed it on a different line from Keycloak.

PR #1087, opened 2025-01-20
support RSA256 with PSS padding Signature alg

kalanthales:

PR #1107, opened 2025-03-15
Add PS256, PS384, and PS512 support

ynojima:
based on #1087 by @kalanthales


ynojima:
Passed all the FIDO conformance test cases including PS256, PS384, and PS512.
image

PR #1208, opened 2025-11-29
Clause Guard approach has been used instead of `else` conditions for more readable codes in `ClientPlatform` class

agitrubard:
This PR refactors the ClientPlatform class to use a clause guard approach instead of nested else blocks, in order to improve readability and maintainability.

Changes
- Replaced wildcard imports with explicit imports for the data, client, and extension.client packages.
  - https://www.baeldung.com/intellij-disable-wildcard-import
- Refactored `create(...)` logic by:
- Extracting `getCollectedClientData(...)` to encapsulate the collected client data override logic.
- Extracting `getAttestationStatement(...)` to handle attestation conveyance preference with clearer early returns.
- Simplified `createCollectedClientData(ClientDataType, Challenge)` by using a guard clause instead of an if/else chain.
- Added `getAuthenticatorAdaptor()` accessor for the internal `authenticatorAdaptor` field.

The behavior of ClientPlatform remains unchanged. All existing tests have been executed and pass successfully.


ynojima:
Thanks for the cleanup — I appreciate the effort. I don't see a clear benefit to merging this refactor: the new private method names are unclear, so I would prefer not to merge as-is.

PR #1337, opened 2026-05-14
Add ML-DSA post-quantum signature algorithm support

### LINKED ISSUE #1200: Post Quantum Cryptography Support
### Discussed in https://github.com/webauthn4j/webauthn4j/discussions/1199

ML-DSA is registered in the COSE algorithms registry

ynojima:
While no standard test vectors or WebAuthn conformance tests for ML-DSA are currently available, this implementation is ready for early testing. If you are interested, you can try it through JitPack.

## Summary

- Add ML-DSA-44/65/87 post-quantum signature algorithms as defined in FIPS 204 and RFC 9964
- Add the AKP (Algorithm Key Pair, `kty=7`) COSE key type for PQC algorithms
- Add ML-DSA to `JWAIdentifier` for JWS/MDS BLOB signature verification

### COSE layer

- `COSEKeyType`: Add `AKP(7)`
- `COSEAlgorithmIdentifier`: Add `ML_DSA_44(-48)`, `ML_DSA_65(-49)`, `ML_DSA_87(-50)`
- `SignatureAlgorithm`: Add `ML_DSA_44`, `ML_DSA_65`, `ML_DSA_87` with JCA names
- `AKPCOSEKey`: New COSE key class for AKP, with DER encoding/decoding for JCA interop and RFC 9964 seed-format private keys
- `COSEKey`: Add a `@JsonSubTypes` entry for AKP

### JWS layer

- `JWAIdentifier`: Add `ML_DSA_44`, `ML_DSA_65`, `ML_DSA_87`
- `JWS`: Return `false` from `isValidSignature()` for unsupported algorithms instead of throwing

### Test authenticator

- `WebAuthnModelAuthenticator`: Add support for ML-DSA key pair generation in `makeCredential()`

### Runtime requirements

- ML-DSA requires JDK 24+ (SUN provider) or Bouncy Castle at runtime.
- The library continues to compile and run on JDK 17.
- ML-DSA tests are gated with `@EnabledIfMLDSAAvailable`, which checks actual algorithm availability rather than JDK version.

### Standards references

- [NIST FIPS 204 (ML-DSA)](https://csrc.nist.gov/pubs/fips/204/final)
- [RFC 9964 (ML-DSA for JOSE and COSE)](https://datatracker.ietf.org/doc/rfc9964/)
- [IANA JOSE registry (JWA algorithm registration)](https://www.iana.org/assignments/jose/jose.xhtml)


Arpan0995:
@ynojima Thanks for the PR — I went through it in depth and ran the full test suite plus some targeted experiments on JDK 24 (Corretto 24.0.2). Overall the structure looks right: the COSE/JOSE identifiers all match the IANA registrations and draft-ietf-cose-dilithium-11 (I verified `-48/-49/-50`, `kty=7`, `pub`/`priv` labels `-1`/`-2`, and the JOSE alg names against the draft), modeling ML-DSA as a pure signature with `messageDigestAlgorithm = null` is correct, the `JWS.isValidSignature` branch on `ECPublicKey` keeps the verify path sound, and `AssertionSignatureVerifier` already catches `RuntimeException`, so the authentication path degrades gracefully on JDK ≤ 23. The full core test suite, including the new AKP/ML-DSA tests, passes for me.

I found a few things worth addressing, roughly in order of importance.

## 1. `priv` holds the expanded private key, but the draft requires the 32-byte seed

draft-ietf-cose-dilithium-11 is explicit:

> For the ML-DSA private keys described in this document, the `priv` parameter MUST be the seed, and MUST have a length of 32 bytes.

JDK 24's PKCS#8 encoding for ML-DSA uses the `expandedKey` CHOICE alternative (an inner OCTET STRING wrapping the expanded key). I confirmed empirically: for ML-DSA-44 the encoded PKCS#8 is 2,588 bytes and `AKPCOSEKey.create(KeyPair, alg)` populates `priv` with the 2,560-byte expanded key, not the 32-byte seed — violating the draft, and the Javadoc ("seed for ML-DSA") doesn't match the behavior. Conversely, a spec-compliant AKP key from another implementation (32-byte seed in `priv`) gets re-wrapped by `buildPKCS8PrivateKeyInfo` as a 32-byte `expandedKey` — and interestingly, `getPrivateKey()` *succeeds* (the JDK's named-key factory doesn't length-check at import); the failure only surfaces later, at `Signature.initSign`, as `InvalidKeyException: Incorrect private key size`.

Since the RP verification path only needs `pub`, one option is to scope `priv` support down (or move it to webauthn4j-test); if `priv` stays in core, it probably needs seed semantics — which is awkward on JDK 24 since `NamedPKCS8Key` doesn't expose the seed. Worth an explicit decision either way.

## 2. Registration-path failure semantics for malformed AKP keys

`CoreRegistrationDataVerifier.verifyCOSEKey` calls `coseKey.getPublicKey()` with no exception mapping, and `AKPCOSEKey.validate()` checks neither the algorithm family nor the key length. Three concrete consequences (all reproduced):

- **Wrong-length `pub` is accepted at registration.** FIPS 204 fixes the `pub` sizes exactly (1,312/1,952/2,592 bytes; the draft repeats them in its own table), and the draft's §7.3 (Validation of keys) states "all algorithm-related key parameters MUST be validated" before use — but a 1,312-byte key declared as ML-DSA-65 passes `validate()` *and* `getPublicKey()`, since the JDK's named-key `KeyFactory` doesn't length-check at import. The RP stores the credential and every subsequent authentication fails (`InvalidKeyException: Incorrect public key size` → `BadSignatureException`). A conformant server should reject this at registration.
- **`kty=7` with a non-ML-DSA `alg` (e.g. ES256) escapes as `UnexpectedCheckedException`** (a bare `RuntimeException`, not a `VerificationException`) from the registration flow, since `KeyFactory.getInstance("SHA256withECDSA")` throws `NoSuchAlgorithmException`. Attacker-controlled input producing an unclassified runtime exception is the kind of thing RPs turn into 500s.
- **On JDK ≤ 23**, any ML-DSA registration attempt takes the same `UnexpectedCheckedException` path out of `verifyCOSEKey`, rather than a typed verification error saying the algorithm is unsupported at runtime.

Suggestion: have `AKPCOSEKey.validate()` enforce (a) `alg` ∈ {ML-DSA-44, -65, -87} and (b) exact `pub` length for the declared parameter set — that converts all three cases into `ConstraintViolationException` at the right ceremony step. This also seems like exactly the negative-test surface the FIDO conformance work (fido-alliance/conformance-test-tools-resources#817) will want: wrong-length keys, cross-parameter-set confusion, truncated signatures.

## 3. Minor: dead branch in `validate()`

The `!hasPublicKey() && !hasPrivateKey()` check is unreachable given the unconditional `pub == null` check right after it — as written, `pub` is always required (which is correct for RP use), so the "pub or priv must be present" branch and message are misleading.

## 4. Provider portability of the hand-rolled PKCS#8 parsing

`extractRawFromPKCS8` assumes the provider encodes `privateKey` as OCTET STRING { OCTET STRING { raw } } (the `expandedKey` alternative). The LAMPS private-key format went through several revisions (seed `[0]`, `expandedKey`, `both`), and providers differ: BouncyCastle's encoding choices have not matched the JDK's across versions, and `KeyFactory.getInstance(alg)` without a provider argument means first-registered-provider-wins — in FIPS deployments (e.g. Keycloak with BC-FIPS) BC commonly sits ahead of SUN in the provider order. If BC returns a seed-based or `both` encoding, the current parse either misextracts or `ClassCastException`s. If `priv` handling moves to the test module per point 1, this mostly goes away; otherwise the parser should handle all three CHOICE alternatives.

## 5. Minor: MDS BLOB path on JDK ≤ 23

`JWS.isValidSignature` catches only `SignatureException | InvalidKeyException`. With ML-DSA now parseable in `JWAIdentifier`, an ML-DSA-signed MDS BLOB processed on JDK ≤ 23 throws `IllegalArgumentException` out of `SignatureUtil.createSignature` (which wraps `NoSuchAlgorithmException`) instead of returning false. Catching the runtime wrapper (or checking algorithm availability first) would keep the JDK 17 baseline clean.

## Test coverage suggestions

The happy-path coverage is good (all three parameter sets, sign/verify, CBOR round-trip, full registration+authentication flow). Gaps worth filling, mostly derived from the above: wrong-length `pub` rejected at registration; `kty`/`alg` mismatch rejected with a `VerificationException`; tampered/truncated assertion signature → `BadSignatureException` (the JDK returns false for tampered sigs of correct length, but truncated ones throw — both paths worth pinning); and packed **self-attestation** with an ML-DSA credential key, which the current integration test doesn't exercise since it registers with attestation `none` and a non-strict manager.

Happy to help with any of these — in particular the seed-vs-expanded-key question and cross-provider (SUN vs BC) behavior, which I've been benchmarking on the JVM.

ynojima:
@Arpan0995 Thank you very much for such a thorough and detailed review. I greatly appreciate the time and effort you put into this.
I'd like to take a few days to carefully consider each of your points. I'll follow up here once I'm ready to respond in detail.

ynojima:
@Arpan0995 Thank you for your patience.

First, I want to apologize that you had to review this in a state where error handling and validation were incomplete. This PR was originally a PoC focused on getting interoperability testing working first, and I should have addressed more of the non-happy-path cases before asking for review. I appreciate you taking the time to go through it so thoroughly despite that.

Your point about the `priv` field holding the expanded private key instead of the 32-byte seed was particularly valuable. This is something I had completely missed, and it wouldn't have surfaced in interoperability testing since only public keys and signatures appear on the wire. Thank you for catching it.

I've addressed the points you raised:

1. **`priv` now stores the 32-byte seed** per RFC 9964. Bouncy Castle is used as an optional dependency to expand the seed back to a private key when needed.
2. **Validation in `AKPCOSEKey.validate()`** now rejects wrong-length `pub`, non-ML-DSA algorithms, and ML-DSA on unsupported JDKs (< 24) with `ConstraintViolationException`.
3. **Dead branch in `validate()`** removed.
4. **`extractRawFromPKCS8`** removed from production code; the seed is now obtained directly via Bouncy Castle.
5. **`JWS.isValidSignature`** now catches `IllegalArgumentException` for unsupported algorithms and returns false.

Test coverage additions:
- Packed self-attestation and basic attestation (x5c) verification tests for ML-DSA
- Bad signature (tampered assertion) test for ML-DSA
- `testWithoutBc` test suite verifying server-side authentication works without Bouncy Castle
- `testOnJDK17` test suite verifying graceful rejection on unsupported JDKs

Arpan0995:
@ynojima Thank you for the follow-up, and no apology needed. A PoC is exactly
the right stage for this, before the wire format and the API settle.

Glad the priv/seed point was useful. I will pull the branch and go through the
new commits properly, in particular the validate() checks, the seed expansion
path, and the two new test suites, and follow up here.

Happy to keep helping on the PQC side going forward. Cross-provider behavior is
something I have been measuring, including a JDK/BouncyCastle/ACCP differential
for ML-DSA, and the negative-test surface for the FIDO conformance work is
another area I have looked at, so if either is useful as this moves ahead, I am
glad to contribute.

Arpan0995:
@ynojima Went through the new commits, against ec9341b0. (I read the core files at 0989020, just before the Spring Security integration commit; they are byte-identical at both.) The five points are addressed, and I verified the seed path end to end rather than reading it. Three findings: a cross-provider result, which was the open question from my point 4, a note on the create(KeyPair) decision, and one small gap.

## Cross-provider: the seed path is portable across SUN and BC

I reproduced `getPrivateKey()` exactly (BC `MLDSAPrivateKeyParameters(params, seed).getEncoded()`, then `buildPKCS8PrivateKeyInfo`, then `KeyFactory.generatePrivate`) on JDK 24.0.2 with BC 1.85, under three provider configurations, and signed with each resulting key:

| provider setup | KeyFactory / Signature resolve to | generatePrivate | initSign + sign |
|---|---|---|---|
| SUN only, BC on classpath but not registered | SUN | OK | OK, 2420 B |
| BC registered last | SUN | OK | OK, 2420 B |
| BC inserted at position 1 (FIPS-style ordering) | BC | OK, `BCMLDSAPrivateKey` | OK, 2420 B |

So the `expandedKey` CHOICE that `buildPKCS8PrivateKeyInfo` emits is accepted by both providers, in both orderings, and the resulting key signs in both. The concern I raised in point 4 does not materialise for BC 1.85. Seed expansion also returns the expanded key rather than the seed, as the JDK import needs: 2560 / 4032 / 4896 bytes for ML-DSA-44 / 65 / 87, giving 2588 / 4060 / 4924 byte PKCS#8, and signatures of 2420 / 3309 / 4627 bytes on all three parameter sets.

Worth recording why this works, because it is not obvious. BC's own JCA private key encodes differently: `BCMLDSAPrivateKey.getEncoded()` produces a 2626-byte PKCS#8 whose `privateKey` is a SEQUENCE of a 32-byte seed and a 2560-byte expanded key, that is the `both` CHOICE, and the JDK rejects it outright with `InvalidKeySpecException: Cannot parse input`. The code here does not go through that path. It calls `MLDSAPrivateKeyParameters.getEncoded()`, the low-level crypto class, which returns the raw 2560-byte expanded key, and then `buildPKCS8PrivateKeyInfo` wraps it as the `expandedKey` CHOICE, which both providers accept. So the portability comes from using the raw parameters object rather than the JCA key, and a refactor that swapped in the JCA `getEncoded()` would look equivalent and break on SUN. Might be worth a comment in `MLDSASeedExpander` so that does not happen later.

One caveat: this is BouncyCastle 1.85, the version the build pins. I have not tested BC-FIPS, which is a separate artifact and the one that actually matters for the Keycloak-style deployments I mentioned. Worth confirming separately if you want to claim FIPS-mode support.

## The create(KeyPair) decision

Dropping `create(KeyPair, alg)` and documenting why, rather than trying to recover a seed the JDK does not expose, looks like the right call, and it is the explicit decision I was asking for in point 1. `create(PublicKey, alg)` covers the RP path, which is the only one that needs to work without Bouncy Castle.

## Minor: no seed-length guard where the seed is consumed

`validate()` enforces `priv.length == 32`, but `MLDSASeedExpander.expand` passes the seed straight to BC without checking. BC throws at construction for any other length, on BC 1.85:

```
seed  0 B -> ArrayIndexOutOfBoundsException: arraycopy: length -32 is negative
seed 16 B -> ArrayIndexOutOfBoundsException: arraycopy: length -16 is negative
seed 31 B -> ArrayIndexOutOfBoundsException: arraycopy: length  -1 is negative
seed 32 B -> OK, 2560 bytes
seed 33 B -> ArrayIndexOutOfBoundsException: arraycopy: length -31 is negative
```

`expandMLDSASeedToPrivateKey` catches `NoClassDefFoundError`, `InvalidKeySpecException` and `NoSuchAlgorithmException`, so this unchecked exception propagates as-is, with an internal arraycopy message.

This is not reachable from the server-side flows: `BeanAssertUtil.validate` runs before `verifyCOSEKey`, and `AKPCOSEKey.getPrivateKey()` is only called from webauthn4j-test (checked across all eleven modules). So it is API hygiene rather than a verification issue.

It is worth a line anyway because removing `create(KeyPair, alg)` makes the constructor the only way to supply a `priv`, and the constructor clones the array without checking it. `validate()` catches a wrong length, but nothing between the constructor and BC does, and `expand` has exactly one caller. It is the same shape as my earlier point 2: unvalidated input leaving as an unclassified runtime exception. A length check in `expand`, or in `getPrivateKey` before calling it, would keep it consistent with how `pub` is treated.

## Everything else checked out

`validate()` is correctly wired into the registration path, `AKPCOSEKeySerializer` matches the `EC2COSEKeySerializer` convention and its label order is correct for CTAP2 canonical CBOR (positive labels before negative), `extractRawFromPKCS8` is gone from production, and the `IllegalArgumentException` catch in `JWS.isValidSignature` is scoped narrowly to `createSignature` rather than the whole verify block, which is the right place for it.

Happy to run the same provider matrix against BC-FIPS if that would be useful.

ynojima:
@Arpan0995 Thank you again for the detailed follow-up.

**SUN/BC compatibility**: Thank you for verifying the seed path across provider configurations. I've added a comment in `MLDSASeedExpander` explaining why `MLDSAPrivateKeyParameters.getEncoded()` (raw expanded key) must be used instead of `BCMLDSAPrivateKey.getEncoded()` (the `both` CHOICE that SUN rejects), along with a test that verifies the expanded key round-trips through the SUN provider's `KeyFactory`.

**Seed-length guard**: Addressed. `getPublicKey()` and `getPrivateKey()` now call `validate()` at the top, so the `priv.length == 32` check runs before the seed reaches `MLDSASeedExpander.expand`. This means an invalid seed length is rejected with `ConstraintViolationException` rather than propagating as an internal `ArrayIndexOutOfBoundsException`.

**Cross-provider testing**: Your mention of BC-FIPS deployments reminded me that WebAuthn4J hadn't had automated testing under non-default provider configurations. I took this as an opportunity to set that up:

- #1478 — `testWithBouncyCastle`: runs all tests with BC registered as the top-priority JCA provider
- #1479 — `testWithBouncyCastleFIPS`: runs all tests with BC-FIPS as the sole JCA provider

I've rebased the ML-DSA branch on top of these changes. With regular BC (1.85), ML-DSA works as expected — the seed expansion path and all ML-DSA tests pass. With BC-FIPS, however, ML-DSA is not yet available. While `bcpqc-fips` (2.1.0) exists as a separate artifact, it provides ML-DSA outside the FIPS certification boundary, so I chose not to depend on it. For now, ML-DSA tests are skipped under BC-FIPS using a custom `@EnabledIfMLDSAAvailable` annotation that checks `KeyFactory.getInstance("ML-DSA-65")` at runtime. Once a future BC-FIPS release brings ML-DSA into the certified boundary, those tests should light up automatically.

Arpan0995:
@ynojima Verified the new commits, against 63a74f87. Everything checks out, and turning the provider question into permanent CI is a good outcome for the project well beyond this PR. One environment note below on the BC-FIPS suite.

## Seed-length guard, verified end to end

`getPrivateKey()` now rejects a wrong-length seed before it reaches Bouncy Castle. I probed through `getPrivateKey()` itself rather than `expand()` directly this time, since the fix is at that call site. JDK 24.0.2 with BC 1.85:

```
seed  0 B -> ConstraintViolationException: ML-DSA priv must be 32 bytes (seed), but was 0
seed 16 B -> ConstraintViolationException: ML-DSA priv must be 32 bytes (seed), but was 16
seed 31 B -> ConstraintViolationException: ML-DSA priv must be 32 bytes (seed), but was 31
seed 33 B -> ConstraintViolationException: ML-DSA priv must be 32 bytes (seed), but was 33
seed 64 B -> ConstraintViolationException: ML-DSA priv must be 32 bytes (seed), but was 64
seed 32 B -> OK, sun.security.pkcs.NamedPKCS8Key, initSign + sign OK (3309 B)
```

Calling `validate()` first also means `getPublicKey()` enforces the pub-length and alg checks on every path, and the comment explaining why the KeyFactory probe is not cached (late provider registration, GraalVM native-image) is a detail I would not have thought to guard against.

## MLDSASeedExpander comment and SUN round-trip test

Both in place. I also checked that the later MLDSAUtil refactor (60770006) only touches test code (webauthn4j-test plus one test class in core), so the raw `MLDSAPrivateKeyParameters.getEncoded()` path in production code is unchanged. `AKPCOSEKeyTest` passes 29/29 for me on JDK 24, including `seed_expansion_produces_sun_compatible_pkcs8_test`, so the invariant is pinned.

## testWithBouncyCastle passes locally

Ran the new suite with BC 1.85 as the primary provider: 1258 tests, 0 failures. Glad to see the one-off matrix I ran by hand now living in the build.

## One note: testWithBouncyCastleFIPS fails on macOS

The suite is wired into `test` via `dependsOn`, and the `onlyIf` guard only excludes Windows, but as the comment in the build file itself says, the JENT native library exists only for Intel/ARM Linux. On macOS (Apple Silicon, JDK 24) the suite therefore runs and fails: 1256 tests, 382 failures. The root cause is the module self-test:

```
org.bouncycastle.crypto.fips.FipsOperationError: Module in error status:
  Exception on self test: java.security.NoSuchAlgorithmException:
  Unable to invoke creator for DEFAULT: JENT RNG failed: EC
```

and since the suite strips the JCA down to BC-FIPS plus the JENT entropy provider, every `SecureRandom` construction after that hits the errored module, down to `NoClassDefFoundError: Could not initialize class java.util.UUID$Holder` from plain `new SecureRandom()` in the JDK itself. CI does not catch it because the matrix runs ubuntu and windows, the guard already skips the suite on windows, and macOS is commented out of the matrix. A plain `./gradlew build` on a Mac now fails though; I reproduced it on both this branch and current master (2827f647). Tightening the guard to match the task's own "(Linux only)" description would fix it:

```kotlin
onlyIf { System.getProperty("os.name").lowercase().contains("linux") }
```

## BC-FIPS boundary

Agree with not depending on bcpqc-fips. ML-DSA outside the certified boundary would defeat the point of a FIPS suite, and the `@EnabledIfMLDSAAvailable` runtime check means the tests light up on their own once a BC-FIPS release brings ML-DSA inside the boundary.

That closes out the question I raised, and from my side everything from both rounds is now addressed. Thanks for working through it all so thoroughly.

ynojima:
@Arpan0995 Thank you for catching the macOS issue with `testWithBouncyCastleFIPS`. Fixed in #1485 — the guard now requires Linux explicitly, and macOS has been added back to the CI matrix.

Arpan0995:
Noticed the rebase today carries one change outside the ML-DSA scope: `JWSSignatureUtil.convertDerSignatureToJwsSignature` now takes the alg and uses a fixed component length (32/48/66) instead of `max(len(r), len(s))`. That reads as a fix for a real master bug rather than POC material, so flagging it in case it is worth landing on its own.

On master the `max()` approach produces a short R||S whenever both r and s happen to fit in fewer bytes than the curve's component size. I measured it with 20,000 signatures per curve (JDK 27 EA, default SunEC, replicating the master logic line for line including the single leading-zero strip in `toUnsignedByteArray`):

- ES256: 1 of 20,000 short (62 bytes instead of 64)
- ES384: 1 of 20,000 short (94 instead of 96)
- ES512: 4,933 of 20,000, about 24.7 percent, short (130 instead of 132)

ES512 is hit that often because the P-521 group order only uses one bit of the top byte, so r and s each fit in 65 bytes roughly half the time. RFC 7518 section 3.4 fixes the sizes at 64/96/132, so a strict verifier rejects these. It went unnoticed internally because `convertJwsSignatureToDerSignature` splits at `length / 2`, which round-trips a 130-byte signature without complaint. To be clear about scope: this is the signing side (`JWSFactory.create`, which is public API); the RP verification path is not affected.

Since this PR is marked not for merge, the fix would otherwise stay stranded here. Two suggestions:

1. Cherry-pick the `JWSSignatureUtil` / `JWSFactory` change to master as its own PR.
2. Add a regression test alongside it. The existing `JWSFactoryTest` signs once with ES256, which has roughly a 1 in 65,000 chance of hitting the case, and does not check the signature length; nothing covers ES512. Either feed a DER signature whose r and s both carry a leading zero byte, or assert the output length equals `2 * componentLength` for each of ES256/ES384/ES512.

One small leftover from the rebase: the class Javadoc on `AKPCOSEKey` (line 50) still links draft-ietf-cose-dilithium, while the comment at line 98 already cites RFC 9964 Section 4. Worth pointing both at the RFC.

Arpan0995:
Saw #1514 land on master, good to have that out of the POC branch. One follow-up on its tests: since each `signature_length_*` test signs a single random key, they only hit the short-component case by chance (about a quarter of the time for ES512, almost never for ES256 and ES384). With the old logic restored locally, `JWSFactoryTest` still passed in 17 of 20 runs, and the ES256 and ES384 tests never failed. I opened #1516 with a deterministic version that builds the DER input with short r and s for each curve, so a regression fails every time. Test-only, no production change.

ynojima:
@Arpan0995 Thank you for the feedback.

**JWS signature length fix**: You're right that this should be a separate PR. While reviewing the commits in this branch, I noticed the ECDSA signature issue and fixed it inline, but it was mixing concerns. I've now split it out into #1514, which has been merged to master. This branch has been rebased to remove those changes.

**Javadoc reference**: Fixed in ff7dc7e2 — both references now point to RFC 9964.

PR #1516, opened 2026-09-02
Add deterministic tests for ECDSA DER to JWS signature conversion

Arpan0995:
Follow-up to #1514.

The three `signature_length_*` tests added there sign a freshly generated key once, so whether they exercise the short-component case depends on the random r and s: roughly a quarter of ES512 signatures, and on the order of 1 in 65,000 for ES256 and ES384. To check, I restored the length-from-max logic locally and ran `JWSFactoryTest` twenty times: the class passed in 17 of the 20 runs. The only failures were `signature_length_es512_test` (3 of 20); the ES256 and ES384 tests never failed.

This adds `JWSSignatureUtilTest`, which builds the DER input with r and s deliberately one byte short of the component size for each curve, so a regression to length-from-max fails on every run. It also covers full-length components, the DER round trip, and the `IllegalArgumentException` for a non-ECDSA algorithm.

Test-only change. Passes under `test` and `testWithBouncyCastle` on JDK 24 (24 JWS tests in total).

### DIFF
--- a/webauthn4j-core/src/test/java/com/webauthn4j/data/jws/JWSSignatureUtilTest.java
+++ b/webauthn4j-core/src/test/java/com/webauthn4j/data/jws/JWSSignatureUtilTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


ynojima:
Good catch! Deterministic tests are preferable.

